PR #6, opened 2021-02-12
RUMM-1075 Use flavor name in the variant tag

xgouchet:
### What does this PR do?

Ensure we use the flavor name and not the variant to match the tag sent by the SDK


mariusc83 (approved):
LGTM

PR #7, opened 2021-02-12
RUMM-1059 Only apply task for minified variants

xgouchet:
### What does this PR do?

Prevent creating upload tasks on non minified variants.

### Motivation

It will prevent having too many gradle tasks in the project, reducing the noise in the tasks list, and avoid creating a task that will do nothing because no corresponding mapping file exists.

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
@@ -64,7 +64,12 @@ class DdAndroidGradlePlugin : Plugin<Project> {
         variant: ApplicationVariant,
         apiKey: String,
         extension: DdExtension
-    ): Task {
+    ): Task? {
+        if (!variant.buildType.isMinifyEnabled) {
+            // Proguard/R8 are not enabled, let's not create a task
+            return null
+        }
+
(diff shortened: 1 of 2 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -64,14 +64,13 @@ class DdAndroidGradlePlugin : Plugin<Project> {
...
         apiKey: String,
         extension: DdExtension
-    ): Task {
-        val flavorName = if (variant.name.endsWith(SUFFIX_DEBUG)) {
-            variant.name.removeSuffix(SUFFIX_DEBUG)
-        } else if (variant.name.endsWith(SUFFIX_RELEASE)) {
-            variant.name.removeSuffix(SUFFIX_RELEASE)
-        } else {
-            variant.name
+    ): Task? {
```
mariusc83, line 67: why do you need to return the Task from here ?

mariusc83 (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
xgouchet: For the tests :)

PR #8, opened 2021-02-18
RUMM-1058 Allow disabling the plugin

xgouchet:
### What does this PR do?

Let customer disable the plugin to only analyse and create variants when necessary. 
Also delays the validation of API Key, allowing developers to have the plugin initialise on computers where the API Key is undefined.


0xnm (approved):

0xnm:
I may miss something on the business logic side, but code-wise lgtm after exploring the source code.

PR #10, opened 2021-02-26
RUMM-1056: Make variants property non-nullable

0xnm:
### What does this PR do?

This change makes variants property non-nullable to avoid force-casting (`variants!!`) in the build script.

### Motivation

Knowledge that force-casting is bad.

### Additional Notes

Solution looks a bit hacky. Better way is to do injection with `ObjectFactory` as described here https://docs.gradle.org/current/userguide/custom_gradle_types.html#collection_types, but the problem is that this method is incubating, so I guess it doesn't guarantee binary compatibility yet (although it is available since Gradle 5.5).

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtension.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtension.kt
@@ -19,15 +19,27 @@ open class DdExtension : DdExtensionConfiguration() {
      */
     var enabled: Boolean = true
 
+    // introduced because cannot use lateinit with `variants` because of name
+    // ambiguity (method vs property). Consider switching to constructor injection
+    // via https://docs.gradle.org/current/javadoc/org/gradle/api/model/ObjectFactory.html#domainObjectContainer-java.lang.Class-
+    // once it is stable
+    private lateinit var _variants: NamedDomainObjectContainer<DdExtensionConfiguration>
+
     /**
      * Container for the variant's configurations.
      */
-    var variants: NamedDomainObjectContainer<DdExtensionConfiguration>? = null
+    var variants: NamedDomainObjectContainer<DdExtensionConfiguration>
+        get() = _variants
+        set(value) {
+            _variants = value
+        }
 
     /**
      * Closure method to create a DSL for variant configurations.
      */
     fun variants(configureClosure: Closure<DdExtensionConfiguration>) {
-        variants?.configure(configureClosure)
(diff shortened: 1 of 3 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtension.kt
```diff
@@ -19,15 +19,27 @@ open class DdExtension : DdExtensionConfiguration() {
...
+        set(value) {
+            _variants = value
+        }
 
     /**
      * Closure method to create a DSL for variant configurations.
      */
     fun variants(configureClosure: Closure<DdExtensionConfiguration>) {
-        variants?.configure(configureClosure)
+        if (::_variants.isInitialized) {
```
0xnm, line 41: not sure that this check is even needed, because "should never happen" ©️ in the plugin lifecycle.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtension.kt
```diff
@@ -19,15 +19,27 @@ open class DdExtension : DdExtensionConfiguration() {
...
+
     /**
      * Container for the variant's configurations.
      */
-    var variants: NamedDomainObjectContainer<DdExtensionConfiguration>? = null
+    var variants: NamedDomainObjectContainer<DdExtensionConfiguration>
+        get() = _variants
+        set(value) {
+            _variants = value
+        }
```
xgouchet, line 35: Do you really need this property here? You're just using the existing `_variant` property without any additional logic.

xgouchet (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtension.kt
0xnm: yes, because I cannot reference just `::variants.isInitialized`, because it creates ambiguity between `variants` property and `variants` method and compiler cannot resolve it.

Something like this doesn't work either:

```
val foo: KProperty<*> = ::variants
if (foo.isInitialized) { }
```

because Kotlin compiler allows `.isInitialized` call only directly. So either need to do such hack to remove ambiguity, or use `ObjectFactory` (which is incubating). Well, there is also an option to use different names for property and method, but not sure if it is the path we would like to take for the plugin API.

xgouchet (approved):

mariusc83 (approved):

PR #13, opened 2021-03-04
RUMM-1010: Warn or fail the build when Datadog SDK is missing in the project dependencies

0xnm:
### What does this PR do?

This changes build a link between Datadog SDK Gradle plugin and Datadog SDK itself: plugin will be able to notify the user if Datadog SDK is missing in the project dependencies (strictly speaking in the runtime classpath). This is achieved by introducing new `checkProjectDependencies` property of the plugin extension, which has 3 options: `none` - take no action if Datadog SDK is missing, `warn` - log a warning if Datadog SDK is missing, `fail` - fail the build if Datadog SDK is missing. This property can be applied either on the global level or per flavor:

```
datadog {
    site = "US"
    checkProjectDependencies = "fail"
    variants {
        fr {
            site = "EU"
            checkProjectDependencies = "none"
        }
    }
}
```

### Motivation

Get a better feedback to the Datadog SDK user if setup is wrong.

### Additional Notes

Check is done at the pre-compilation step (lazy way), to save build script configuration time which may be large for the big projects since we need to check full dependency tree (not necessarily Datadog SDK will be top level dependency for the project, it may also come from the other modules).

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ datadog {
     versionName = "1.3.0" // Optional, by default it is read from your Android plugin configuration's version name
     serviceName = "my-service" // Optional, by default it is read from your Android plugin configuration's package name
     site = "US" // Optional, can be "US", "EU" or "GOV". Default is "US"
+    checkProjectDependencies = "warn" // Optional, can be "warn", "fail" or "none". Default is "fail". Will check if Datadog SDK is in the project dependencies.
 }
 ```
 
@@ -40,6 +41,7 @@ datadog {
     variants {
         fr {
             site = "EU"
+            checkProjectDependencies = "warn"
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -41,7 +41,6 @@ datadog {
     variants {
         fr {
             site = "EU"
-            checkProjectDependencies = "warn"
         }
     }
 }
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
@@ -9,10 +9,12 @@ package com.datadog.gradle.plugin
 import com.android.build.gradle.AppExtension
 import com.android.build.gradle.api.ApplicationVariant
 import com.datadog.gradle.plugin.internal.GitRepositoryDetector
+import com.datadog.gradle.plugin.internal.MissingSdkException
 import java.io.File
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.artifacts.ResolvedDependency
 import org.slf4j.LoggerFactory
 
 /**
@@ -26,7 +28,7 @@ class DdAndroidGradlePlugin : Plugin<Project> {
     override fun apply(target: Project) {
         val androidExtension = target.extensions.findByType(AppExtension::class.java)
         if (androidExtension == null) {
-            System.err.println(ERROR_NOT_ANDROID)
+            LOGGER.error(ERROR_NOT_ANDROID)
             return
         }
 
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -40,6 +41,7 @@ datadog {
     variants {
         fr {
             site = "EU"
+            checkProjectDependencies = "warn"
```
xgouchet, line 44: Let's not put this in the code snippet, as it will be copy-pasted 90% of the time. The goal of this is to fail at least one if need be.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -26,7 +28,7 @@ class DdAndroidGradlePlugin : Plugin<Project> {
     override fun apply(target: Project) {
         val androidExtension = target.extensions.findByType(AppExtension::class.java)
         if (androidExtension == null) {
-            System.err.println(ERROR_NOT_ANDROID)
+            LOGGER.error(ERROR_NOT_ANDROID)
```
xgouchet, line 31: Good call 👍

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -101,6 +104,69 @@ class DdAndroidGradlePlugin : Plugin<Project> {
...
+
+        // variants will have name like proDebug, but compile task will have a name like
+        // compileProDebugSources. It can be other tasks like compileProDebugAndroidTestSources
+        // or compileProDebugUnitTestSources, but we are not interested in these. This is fragile
+        // and depends on the AGP naming convention
+        val compileTask = target.tasks
+            .findByName("compile${variant.name.capitalize()}Sources")
+
+        if (compileTask == null) {
+            LOGGER.warn(
```
xgouchet, line 126: Maybe we could use the same level of warning for this step as for the actual check, WDYT?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
```diff
@@ -360,6 +375,304 @@ internal class DdAndroidGradlePluginTest {
...
+            testedPlugin.configureVariantForSdkCheck(
+                fakeProject,
+                mockVariant,
+                fakeExtension
+            )!!.actions.first().execute(fakeCompileTask)
+        }
+    }
+
+    @Test
+    fun `𝕄 do nothing 𝕎 configureVariantForSdkCheck() { sdk is there }`(
```
xgouchet, line 520: You're missing a `𝕄 do nothing 𝕎 configureVariantForSdkCheck() { sdk is missing w check = none }`

xgouchet (changes requested):
Really nice work, just a few questions, and you're missing a UT

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
0xnm: I wouldn't link it to the actual check, actually this branch exists because `findByName("compile${variant.name.capitalize()}Sources")` above is kind of fragile - it relies on the naming convention introduced by the current version of AGP, which may change in the future versions. So I was thinking about receiving feedback from the users who use some `alpha/beta` versions of AGP where this convention _may be_ different. In theory this "should never happen" ©️ , because then AGP will break backward compatibility with a bunch of other plugins I think. Do you think users will miss this warning?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
0xnm: indeed, good catch!

0xnm:
FYI: I checked setup when DD SDK is declared in another module (other than plugin, lib module) and this module is referenced from app module (plugin module) - everything works fine as well.

xgouchet (approved):

mariusc83 (approved):

PR #14, opened 2021-03-05
RUMM-1127 Fail the build task if mapping file size exceeds the 50 MB …

mariusc83:
### What does this PR do?

In this PR we will check and fail the `uploadMapping` task if the `mappingFile.txt` size exceeds the 50 MB max size.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
@@ -66,7 +66,12 @@ internal class OkHttpUploader : Uploader {
         repositoryInfo: RepositoryInfo?
     ): MultipartBody {
         val mappingFileBody = MultipartBody.create(MEDIA_TYPE_TXT, mappingFile)
-
+        if (mappingFileBody.contentLength() > MAX_MAP_FILE_SIZE_IN_BYTES) {
+            throw MaxSizeExceededException(
+                MAX_MAP_SIZE_EXCEEDED_ERROR_FORMAT
+                    .format(mappingFile.absolutePath)
+            )
+        }
         val builder = MultipartBody.Builder()
         builder.setType(MultipartBody.FORM)
             .addFormDataPart("version", identifier.version)
@@ -118,7 +123,10 @@ internal class OkHttpUploader : Uploader {
         internal val NETWORK_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(45)
         internal val MEDIA_TYPE_TXT = MediaType.parse("text/plain")
         internal val MEDIA_TYPE_JSON = MediaType.parse("application/json")
-
+        internal const val MAX_MAP_FILE_SIZE_IN_BYTES = 50L * 1024L * 1024L // 50 MB
+        internal val MAX_MAP_SIZE_EXCEEDED_ERROR_FORMAT =
+            "The proguard mapping file at: [%s] size exceeded the maximum 50 MB size. " +
+                "This task cannot be performed."
         internal val succesfulCodes = arrayOf(
             HttpURLConnection.HTTP_OK,
             HttpURLConnection.HTTP_CREATED,
--- a/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploaderTest.kt
+++ b/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploaderTest.kt
@@ -9,6 +9,10 @@ package com.datadog.gradle.plugin.internal
 import com.datadog.gradle.plugin.Configurator
 import com.datadog.gradle.plugin.RecordedRequestAssert.Companion.assertThat
 import com.datadog.gradle.plugin.RepositoryInfo
+import com.nhaarman.mockitokotlin2.doReturn
+import com.nhaarman.mockitokotlin2.spy
+import com.nhaarman.mockitokotlin2.whenever
+import fr.xgouchet.elmyr.Forge
 import fr.xgouchet.elmyr.annotation.Forgery
 import fr.xgouchet.elmyr.annotation.IntForgery
 import fr.xgouchet.elmyr.annotation.StringForgery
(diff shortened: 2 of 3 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
```diff
@@ -111,14 +117,22 @@ internal class OkHttpUploader : Uploader {
         }
     }
 
+    private fun RequestBody.sizeInMb(): Long {
+        val megabyte = 1024L * 1024L
```
mariusc83, line 121: Actually here I need to make a round to the biggest long otherwise we can have sizes of `50.x` that will be considered valid.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
```diff
@@ -66,7 +67,12 @@ internal class OkHttpUploader : Uploader {
         repositoryInfo: RepositoryInfo?
     ): MultipartBody {
         val mappingFileBody = MultipartBody.create(MEDIA_TYPE_TXT, mappingFile)
-
+        if (mappingFileBody.sizeInMb() > MAX_MAP_FILE_SIZE_IN_MB) {
```
0xnm, line 70: maybe weird question, but what if file size is 50 MB and 1 byte? it is also not clear from the task if it is inclusive or exclusive bound. maybe we can check that in backend code?

0xnm (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
mariusc83: check the unit tests...there's a specific test there for that.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
```diff
@@ -111,14 +117,22 @@ internal class OkHttpUploader : Uploader {
...
+
     // endregion
 
     companion object {
 
         internal val NETWORK_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(45)
         internal val MEDIA_TYPE_TXT = MediaType.parse("text/plain")
         internal val MEDIA_TYPE_JSON = MediaType.parse("application/json")
-
+        private const val MAX_MAP_FILE_SIZE_IN_MB = 50.0
```
xgouchet, line 132: Why is this value a double and not a long ? Wouldn't it be easier to keep a 
`MAX_MAP_FILE_SIZE_IN_MB = 50L * 1024L * 1024L` and compare that to the actual size of the file?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploaderTest.kt
```diff
@@ -275,6 +279,105 @@ internal class OkHttpUploaderTest {
...
+                    fakeRepositoryFile,
+                    fakeIdentifier,
+                    fakeRepositoryInfo
+                )
+            }
+        )
+    }
+
+    @Test
+    fun `M throw a MaxSizeExceededException W upload() { mappingFile size between 50 and 51 MB}`(
```
xgouchet, line 306: Technically a file size between 50 and 51MB is always greater than or equal to 50MB. I don't really see the point of having this additional test. I guess we can have three tests: size<50Mb, size==50Mb and size>50Mb.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploaderTest.kt
```diff
@@ -275,6 +279,105 @@ internal class OkHttpUploaderTest {
...
+    }
+
+    @Test
+    fun `M fail and do nothing W upload() { mappingFile size exceeded 50 MB }`(
+        forge: Forge
+    ) {
+        // GIVEN
+        val fakeTooLargeMappingFile = spy(File(tempDir, forge.anAlphabeticalString()))
+        doReturn(forge.aLong(min = 51))
+            .whenever(fakeTooLargeMappingFile).length()
```
xgouchet, line 337: Aren't you faking a file with size ≥ 51 bytes? How can you be sure that it exceeds 50Mb ? Also isn't this redundant with the test above?

xgouchet (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
mariusc83: Yes....I really don't know why I did it the other way around :)

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploaderTest.kt
mariusc83: Oh...sorry I forgot to push a new commit where I cleaned up these tests. I will change though the logic following your suggestion and as it is much more simpler and will addd these 3 tests.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploaderTest.kt
```diff
@@ -275,6 +279,111 @@ internal class OkHttpUploaderTest {
...
+                    fakeRepositoryFile,
+                    fakeIdentifier,
+                    fakeRepositoryInfo
+                )
+            }
+        )
+    }
+
+    @Test
+    fun `M fail and do nothing W upload() { mappingFile size exceeded 50 MB }`(
```
0xnm, line 308: seems like this test is equal to the test above and the difference is just additional `assertThat(mockWebServer.requestCount).isEqualTo(0)` assertion. maybe this test can be removed and assertion merged with test above?

mariusc83: yes...we do this intentionally as we want to have 1 test/assertion. This way if your test fails in CI you know exactly what it  failed (either sending a log message, either skipping the writer call)

xgouchet: I do agree with Nikita, you're duplicating the assertThrows, and simply adding one assertion. Either make it different (only assert on the `mockWebServer.requestCount` here and on the exception above), or combine the two

xgouchet (changes requested):

xgouchet (approved):

PR #15, opened 2021-03-05
RUMM-1103 handle partial variant configurations

xgouchet:
### What does this PR do?

Let customer define plugin configuration with partial variant names. 

### Additional Notes

This uses the same priority as the one defined in the Android compiler when resolving Source Sets.

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
@@ -190,7 +190,8 @@ class DdAndroidGradlePlugin : Plugin<Project> {
         configuration.updateWith(extension)
 
         val flavors = variant.productFlavors.map { it.name }
-        val iterator = VariantIterator(flavors)
+        val buildType = variant.buildType.name
+        val iterator = VariantIterator(flavors + buildType)
         iterator.forEach {
             val config = extension.variants.findByName(it)
             if (config != null) {
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
@@ -10,6 +10,7 @@ import com.android.build.gradle.AppExtension
 import com.android.build.gradle.api.ApplicationVariant
 import com.datadog.gradle.plugin.internal.GitRepositoryDetector
 import com.datadog.gradle.plugin.internal.MissingSdkException
+import com.datadog.gradle.plugin.internal.VariantIterator
 import java.io.File
 import org.gradle.api.Plugin
 import org.gradle.api.Project
@@ -82,7 +83,7 @@ class DdAndroidGradlePlugin : Plugin<Project> {
             DdMappingFileUploadTask::class.java,
             GitRepositoryDetector()
         )
-        val extensionConfiguration = resolveExtensionConfiguration(extension, flavorName)
+        val extensionConfiguration = resolveExtensionConfiguration(extension, variant)
 
         configureVariantTask(uploadTask, apiKey, flavorName, extensionConfiguration, variant)
 
@@ -130,7 +131,6 @@ class DdAndroidGradlePlugin : Plugin<Project> {
             )
             return null
         } else {
-
             // this will postpone the check until the actual compilation for the particular variant.
             // by doing this we are avoiding pulling all the configurations for all the variants
             // after build script is evaluated, which may be a heavy task for the big projects
@@ -143,7 +143,7 @@ class DdAndroidGradlePlugin : Plugin<Project> {
 
                     val extensionConfiguration = resolveExtensionConfiguration(
                         extension,
-                        variant.flavorName
+                        variant
                     )
 
                     val sdkCheckLevel = extensionConfiguration.checkProjectDependencies
@@ -184,17 +184,20 @@ class DdAndroidGradlePlugin : Plugin<Project> {
 
     internal fun resolveExtensionConfiguration(
         extension: DdExtension,
-        flavorName: String
+        variant: ApplicationVariant
     ): DdExtensionConfiguration {
-        val flavorConfig = extension.variants.findByName(flavorName)
+        val configuration = DdExtensionConfiguration()
+        configuration.updateWith(extension)
 
-        return DdExtensionConfiguration().apply {
-            versionName = flavorConfig?.versionName ?: extension.versionName
-            serviceName = flavorConfig?.serviceName ?: extension.serviceName
-            site = flavorConfig?.site ?: extension.site
-            checkProjectDependencies = flavorConfig?.checkProjectDependencies
-                ?: extension.checkProjectDependencies
+        val flavors = variant.productFlavors.map { it.name }
+        val iterator = VariantIterator(flavors)
+        iterator.forEach {
+            val config = extension.variants.findByName(it)
+            if (config != null) {
+                configuration.updateWith(config)
+            }
         }
+        return configuration
     }
 
     internal fun isDatadogDependencyPresent(dependencies: Set<ResolvedDependency>): Boolean {
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/VariantIterator.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/VariantIterator.kt
@@ -1,16 +1,16 @@
 package com.datadog.gradle.plugin.internal
 
 /**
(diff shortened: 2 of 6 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -184,17 +184,20 @@ class DdAndroidGradlePlugin : Plugin<Project> {
...
-            checkProjectDependencies = flavorConfig?.checkProjectDependencies
-                ?: extension.checkProjectDependencies
+        val configuration = DdExtensionConfiguration()
+        configuration.updateWith(extension)
+
+        val flavors = variant.productFlavors.map { it.name }
+        val iterator = VariantIterator(flavors)
+        iterator.forEach {
+            val config = extension.variants.findByName(it)
+            if (config != null) {
```
0xnm, line 196: I think it is useful to run also the opposite check - check entries defined in `extension.variants` which are missing in flavors/variants and show a warning if it is a case. this may happen because of typo, for example, and may go unnoticed for a very long time.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -184,17 +184,20 @@ class DdAndroidGradlePlugin : Plugin<Project> {
...
-            site = flavorConfig?.site ?: extension.site
-            checkProjectDependencies = flavorConfig?.checkProjectDependencies
-                ?: extension.checkProjectDependencies
+        val configuration = DdExtensionConfiguration()
+        configuration.updateWith(extension)
+
+        val flavors = variant.productFlavors.map { it.name }
+        val iterator = VariantIterator(flavors)
+        iterator.forEach {
+            val config = extension.variants.findByName(it)
```
0xnm, line 195: this won't match the case of `buildFreeDebug` (which is mentioned in the related task), because `debug` is not a part of flavor, but of build type.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/VariantIterator.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    override fun hasNext(): Boolean {
+        return index < max
+    }
+
+    override fun next(): String {
+        if (index >= max) {
+            throw NoSuchElementException()
+        }
+        val mask = index.toString(2).padStart(names.size, '0').map { it == '1' }
```
0xnm, line 30: nit: I would suggest to add explanation/example of evaluation as a comment

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
```diff
@@ -710,5 +719,84 @@ internal class DdAndroidGradlePluginTest {
...
+        fakeExtension.variants = mock()
+        whenever(fakeExtension.variants.findByName(flavorA)) doReturn variantConfigA
+        whenever(fakeExtension.variants.findByName(flavorB)) doReturn variantConfigB
+        whenever(fakeExtension.variants.findByName(flavorC)) doReturn variantConfigC
+
+        // When
+        val config = testedPlugin.resolveExtensionConfiguration(fakeExtension, mockVariant)
+
+        // Then
+        assertThat(config.versionName).isEqualTo(variantConfigB.versionName)
```
0xnm, line 745: just a thought: we are giving user a flexible way to define the things, but there is a possibility to shoot in the leg as well. Let's suppose that we have 2 flavors of different dimensions - `foo` and `bar`, and user wrote the following config:

```
datadog {
    site = "US"
    foo {
        serviceName = "foo"
    }
    bar {
        serviceName = "bar"
    }
}
```

In this case variant named `fooBarDebug` will get `serviceName = bar`, but maybe user would like to use `serviceName = foo`. It is a bit weird example, but I'm thinking about big enterprise apps which may have quite a lot of flavors/variants (especially if they are generated dynamically and not directly declared in the build script). Maybe it makes sense to add a check if extension field was updated more than once? wdyt?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/VariantIteratorTest.kt
```diff
@@ -0,0 +1,140 @@
...
+                a + B + C + D
+            )
+    }
+
+    @Test
+    fun `𝕄 iterate 𝕎 forEach() { anySize }`(
+        @StringForgery(case = Case.LOWER) names: List<String>
+    ) {
+        // Given
+        val limitedNames = names.take(10) // limit the complexity of the data
```
0xnm, line 111: that is a good point. `1 << names.size` can grow quickly, especially for big apps and can slow down (or even break) the build. wondering how to find the information how many dimensions can exist in the big apps

0xnm (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/VariantIterator.kt
mariusc83: Or maybe a simplified version:
```
internal class VariantIterator(
    private val names: List<String>
) : Iterator<String> {

    private var index = 0
    private var variants: MutableList<String> = mutableListOf()

    init {
        addVariant("")
    }

    private fun addVariant(prefix: String = "", level: Int = 0) {
        for (i in level until names.size) {
            val variant = prefix + if (level != 0) {
                names[i].capitalize()
            } else {
                names[i]
            }
            variants.add(variant)
            addVariant(variant, i + 1)
        }
    }

    override fun hasNext(): Boolean {
        return index < variants.size
    }

    override fun next(): String {
        if (index >= variants.size) {
            throw NoSuchElementException()
        }
        return variants[index++]
    }
}
```

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/VariantIterator.kt
```diff
@@ -0,0 +1,35 @@
...
+    private var index = 1
+    private val max = 1 shl names.size
+
+    override fun hasNext(): Boolean {
+        return index < max
+    }
+
+    override fun next(): String {
+        if (index >= max) {
+            throw NoSuchElementException()
```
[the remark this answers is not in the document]
xgouchet: No because the underlying data cannot be modified. The contract for `Iterator` specifically says that if `next()` is called but `hasNext()` would return false, the implementation should throw a `NoSuchElementException`

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/VariantIterator.kt
xgouchet: Nice trick Marius, but the idea of this implementation was to avoid generating and storing the full list of all possible combinations (which would be exponential). Also the implementation you gave doesn't provide the correct order of all combinations.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/VariantIteratorTest.kt
xgouchet: I have a guess that even for big apps, the number of flavor dimensions should stay reasonable as it will also impact gradle's build time (not to mention the complexity for engineers to keep track of all those dimensions in their mind when writing their app)

0xnm (approved):
nice work!

mariusc83 (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/VariantIterator.kt
```diff
@@ -0,0 +1,54 @@
...
+        // Because the first name has a higher priority, the sequence would be (for 3 names):
+        // false-false-true, false-true-false, false-true-true, true-false-false, and so on…
+        // This sequence matches exactly the binary representation of integers starting from 1, so:
+        // - the iterator iterates from 1 to (2^n - 1)
+        // - at each step, we generate the binary representation of the current index
+        // - we pad the binary with '0' to match the names size
+        // - we convert the  binary represnetation string into a boolean array
+        // - we zip the array with the name
+        val mask = index.toString(2).padStart(names.size, '0').map { it == '1' }
+        val filteredNames = mask.zip(names) { b, s -> if (b) s else null }.filterNotNull()
```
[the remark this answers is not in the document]
xgouchet: `b`  is for boolean ;)

PR #17, opened 2021-03-11
RUMM-1036: Enable codecov integration

0xnm:


0xnm:
looks like finally it works, ready for the review.

buildSrc/src/main/kotlin/com/datadog/gradle/config/JacocoConfig.kt
```diff
@@ -9,77 +9,58 @@ package com.datadog.gradle.config
...
 
-        task.classDirectories.setFrom(files(debugTree, kotlinDebugTree))
-        task.executionData.setFrom(files("${buildDir.path}/jacoco/testDebugUnitTest.exec"))
+        task.executionData.setFrom(files("${buildDir.path}/jacoco/test.exec"))
         task.sourceDirectories.setFrom(files(mainSrc))
     }
-    jacocoTestDebugUnitTestReport.dependsOn("testDebugUnitTest")
-    jacocoTestReleaseUnitTestReport.dependsOn("testReleaseUnitTest")
-    jacocoTestCoverageVerification.dependsOn(jacocoTestDebugUnitTestReport)
+    jacocoTestReport.dependsOn("test")
```
mariusc83, line 47: here you are making the `jacocoTestReport` to depend on `test` but bellow you actually finalize the `test` task with `jacocoTestReport` . Isn't this going to create a circular dependency ?

mariusc83 (changes requested):

buildSrc/src/main/kotlin/com/datadog/gradle/config/JacocoConfig.kt
0xnm: no, it is fine. `dependsOn` just means that this task should be executed by the time parent task is triggered.

Gradle's DAG is resolved fine because in gradle task name is unique, so if we run `jacocoTestReport` it will run `test` and by the completion of `test` gradle knows that `jacocoTestReport` is in the pending state and its dependencies just completed and should not be run again.

mariusc83 (approved):

xgouchet (approved):

PR #16, opened 2021-03-09
Gradle 6.8.3 & Android Gradle Plugin 4.1.2

0xnm:
### What does this PR do?

This change updates Gradle version to 6.8.3 and Android Gradle Plugin version to 4.1.2 (should be safe to use since it exists for a quite long time already)

### Additional Notes

Will probably open another PR for the Kotlin language update. We can do it safely because there is no interop with client code.

PR #18, opened 2021-03-11
Android Gradle Plugin 4.1.2 in buildSrc module

0xnm:
Follow-up on #16 : one more place needs to be updated.


mariusc83 (approved):
I just noticed that `buildSrc` project overrides the Kotlin compiler version in our projects, from there the deprecation warning on all our `format` methods. Basically all our project are built with the Kotlin compiler declared in `buildSrc` :(.
Do you think we can do any trick to actually use `1.4.0` in buildSrc?

0xnm:
@mariusc83 interesting point, I'm going to take a look

xgouchet (approved):

PR #19, opened 2021-03-15
Use Kotlin stdlib8 instead of plain stdlib

0xnm:
### What does this PR do?

This change resolves the following compilation warning:

```
w: Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
    /Users/<user>/.gradle/wrapper/dists/gradle-6.8.3-all/6czipnbiesy2sl92ioo8dht91/gradle-6.8.3/lib/kotlin-stdlib-1.4.20.jar (version 1.4)
    /Users/<user>/.gradle/wrapper/dists/gradle-6.8.3-all/6czipnbiesy2sl92ioo8dht91/gradle-6.8.3/lib/kotlin-stdlib-common-1.4.20.jar (version 1.4)
    /Users/<user>/.gradle/wrapper/dists/gradle-6.8.3-all/6czipnbiesy2sl92ioo8dht91/gradle-6.8.3/lib/kotlin-stdlib-jdk7-1.4.20.jar (version 1.4)
    /Users/<user>/.gradle/wrapper/dists/gradle-6.8.3-all/6czipnbiesy2sl92ioo8dht91/gradle-6.8.3/lib/kotlin-stdlib-jdk8-1.4.20.jar (version 1.4)
    /Users/<user>/.gradle/wrapper/dists/gradle-6.8.3-all/6czipnbiesy2sl92ioo8dht91/gradle-6.8.3/lib/kotlin-reflect-1.4.20.jar (version 1.4)
    /Users/<user>/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-reflect/1.4.0/aa2101a19d8688e368ae6e35e8967550ced73884/kotlin-reflect-1.4.0.jar (version 1.4)
    /Users/<user>/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-jdk8/1.3.72/916d54b9eb6442b615e6f1488978f551c0674720/kotlin-stdlib-jdk8-1.3.72.jar (version 1.3)
    /Users/<user>/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-jdk7/1.3.72/3adfc2f4ea4243e01204be8081fe63bde6b12815/kotlin-stdlib-jdk7-1.3.72.jar (version 1.3)
[...]

PR #20, opened 2021-03-15
Use embedded Kotlin for the build script, bump Kotlin version to 1.4.20

0xnm:
### What does this PR do?

Better version of #19. It takes advantage of Gradle 6.8.3 shipped with embedded Kotlin version 1.4.20 (originally update was done here https://github.com/gradle/gradle/commit/b5f89f1a7c1accc84806751c2c8941fae43cd33a to 1.4.10 and then also here https://github.com/gradle/gradle/commit/6f634d19a5bf4cb21af1ed9b2bffe3089a33efc3 to 1.4.20).

With Kotlin 1.4 there is no need to declare `stdlib` explicitly, it is added automatically: https://kotlinlang.org/docs/gradle.html#dependency-on-the-standard-library (so I removed it in the `buildSrc`, but kept in the individual build scripts, just in case).

More on embedded Kotlin: https://docs.gradle.org/current/userguide/kotlin_dsl.html#sec:kotlin

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -13,9 +13,6 @@ plugins {
 }
 
 buildscript {
-    dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.61")
-    }
     repositories {
         mavenCentral()
     }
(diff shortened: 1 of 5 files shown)


mariusc83 (approved):
Nicely done !

xgouchet (approved):

PR #24, opened 2021-04-09
RUMM-1115: Prevent listing git ignored files

0xnm:
### What does this PR do?

This change prevents listing patterns mentioned in `.gitignore` by switching from manual folder traversal to using [git ls-files](https://git-scm.com/docs/git-ls-files) which respects this rule.

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
+++ b/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
@@ -12,7 +12,6 @@ import org.assertj.core.api.Assertions.assertThat
 import org.gradle.api.Project
 import org.gradle.testfixtures.ProjectBuilder
 import org.junit.jupiter.api.BeforeEach
-import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.extension.Extensions
@@ -54,7 +53,6 @@ internal class GitRepositoryDetectorTest {
         testedDetector = GitRepositoryDetector()
     }
 
-    @Disabled("This doesn't perform well in the CI, but works locally")
     @Test
     fun `𝕄 return repository info 𝕎 detectRepository()`() {
         // Given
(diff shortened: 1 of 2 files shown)


dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
```diff
@@ -54,7 +53,7 @@ internal class GitRepositoryDetectorTest {
         testedDetector = GitRepositoryDetector()
     }
 
-    @Disabled("This doesn't perform well in the CI, but works locally")
+    // @Disabled("This doesn't perform well in the CI, but works locally")
```
mariusc83, line 56: ??

mariusc83 (changes requested):

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
0xnm: I uncommented it to see what happens on CI (in local everything is fine indeed). Going to pull the image locally and investigate.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
```diff
@@ -109,25 +107,42 @@ internal class GitRepositoryDetectorTest {
...
+                .waitForSuccess(5, TimeUnit.SECONDS)
         )
         check(
             ProcessBuilder("git", "add", ".")
                 .directory(tempDir)
                 .start()
-                .waitFor(5, TimeUnit.SECONDS)
+                .waitForSuccess(5, TimeUnit.SECONDS)
+        )
+        check(
```
0xnm, line 118: so the thing is that CI image has no git user set up, so doing commit there fails. and `waitFor` return `true` only if there is no timeout, it doesn't care about process return value.

mariusc83: So this means in the CI it will fail ? The way I see your new `waitForSuccess` function returns : `haveNotTimeout&&exitValue()==0` . The `exitValue` will be `1` for CI right ?

0xnm: no, I fixed the issue related to the run on CI :) it was related to the missing git user (fixed with `git config` calls)

mariusc83: Oh I see now...nicely done  👍

mariusc83 (approved):

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
```diff
@@ -109,25 +107,42 @@ internal class GitRepositoryDetectorTest {
...
+            ProcessBuilder("git", "config", "user.name", "\"Bruce Wayne\"")
+                .directory(tempDir)
+                .start()
+                .waitForSuccess(5, TimeUnit.SECONDS)
+        )
+        check(
+            ProcessBuilder("git", "config", "user.email", "\"bruce@wayne.com\"")
```
xgouchet, line 125: This might actually be a real adress… https://wayne.com/
Maybe let's use something a bit more generic (john smith @ example .com) for a public repository, even if it's less fun

xgouchet (changes requested):
Really nice work here :)

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
0xnm: done

xgouchet (approved):

PR #25, opened 2021-04-15
Vendor the codecov upload script

xgouchet:
### What does this PR do?

Ensure we don't run in the CI a script downloaded from the internet.

### Motivation

[Codecov security breach](https://about.codecov.io/security-update/)


0xnm (approved):
There is a way to validate the script https://docs.codecov.io/docs/about-the-codecov-bash-uploader#validating-the-bash-script. Does this one match the expected?

xgouchet:
> There is a way to validate the script https://docs.codecov.io/docs/about-the-codecov-bash-uploader#validating-the-bash-script. Does this one match the expected?

Yes I checked it with all three checksum provided :)

PR #26, opened 2021-04-20
Fix POM file missing dependencies

xgouchet:
The Maven plugin creates a publication automatically for jar outputs with the relevant dependencies.
Instead of creating a "blank" publication, we can reuse the existing one and add the missing info there.

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
@@ -20,7 +20,7 @@ object MavenConfig {
 
     val VERSION = Version(1, 0, 0)
     const val GROUP_ID = "com.datadoghq"
-    const val PUBLICATION = "release"
+    const val PUBLICATION = "pluginMaven"
 }
 
 @Suppress("UnstableApiUsage")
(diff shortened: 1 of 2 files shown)


mariusc83 (approved):

buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
```diff
@@ -65,8 +65,8 @@ fun Project.publishingConfig(projectDescription: String) {
                 }
             }
 
-            publications.create(MavenConfig.PUBLICATION, MavenPublication::class.java) {
-                artifact(tasks.findByName("jar"))
+            publications.getByName(MavenConfig.PUBLICATION) {
```
0xnm, line 68: I'm not sure I got it: which piece of code then creates `pluginMaven` publication? because I see, that `getByName` will fail if there is no such object, but I don't see where we create that publication.

Also there is a gradle task named `publishReleasePublicationToMavenRepository` which is run in CI, should it be renamed too then?

xgouchet: Regarding the name, yes it's the main publication [created by default](https://github.com/gradle/gradle/blob/5ec3f672ed600a86280be490395d70b7bc634862/subprojects/plugin-development/src/main/java/org/gradle/plugin/devel/plugins/MavenPluginPublishPlugin.java#L73).

Regarding the Gitlab job, good point :)

0xnm (approved):

PR #37, opened 2021-06-14
RUMM-1411 Remove credentials if any from GIT remote URL

mariusc83:
### What does this PR do?

There were reported situations when the client was using an `HTTPS` Git Remote URL which contained also the `username` and `password`: `https://username:password@github.com/username/repository.git`. In this case we were uploading the full URL together with the `mapping.txt` file to our DD servers.

In this PR we are sanitizing the URL and we remove the user information if any.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
@@ -3,17 +3,18 @@ package com.datadog.gradle.plugin.internal
 import com.datadog.gradle.plugin.DdAndroidGradlePlugin.Companion.LOGGER
 import com.datadog.gradle.plugin.RepositoryDetector
 import com.datadog.gradle.plugin.RepositoryInfo
+import com.datadog.gradle.plugin.internal.sanitizer.GitRemoteUrlSanitizer
+import com.datadog.gradle.plugin.internal.sanitizer.UrlSanitizer
 import java.io.File
-import javax.inject.Inject
 import org.gradle.process.ExecOperations
 import org.gradle.process.internal.ExecException
 
 // TODO RUMM-1095 handle git submodules
 // TODO RUMM-1096 handle git subtrees
 // TODO RUMM-1093 let customer override `origin` with custom remote name
-internal open class GitRepositoryDetector
-@Inject constructor(
-    @Suppress("UnstableApiUsage") private val execOperations: ExecOperations
+internal class GitRepositoryDetector(
+    @Suppress("UnstableApiUsage") private val execOperations: ExecOperations,
+    private val urlSanitizer: UrlSanitizer = GitRemoteUrlSanitizer()
 ) : RepositoryDetector {
 
     @Suppress("StringLiteralDuplication")
@@ -27,7 +28,9 @@ internal open class GitRepositoryDetector
             return emptyList()
         }
 
-        val remoteUrl = execOperations.execShell("git", "remote", "get-url", "origin").trim()
+        val remoteUrl = sanitizeUrl(
+            execOperations.execShell("git", "remote", "get-url", "origin").trim()
+        )
         val commitHash = execOperations.execShell("git", "rev-parse", "HEAD").trim()
 
         val trackedFiles = listTrackedFilesPath(sourceSetRoots)
@@ -42,6 +45,9 @@ internal open class GitRepositoryDetector
     }
 
     // region Internal
+    private fun sanitizeUrl(remoteUrl: String): String {
+        return urlSanitizer.sanitize(remoteUrl)
+    }
 
     private fun listTrackedFilesPath(
         sourceSetRoots: List<File>
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/ShortUriGitRemoteUrlSanitizer.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/ShortUriGitRemoteUrlSanitizer.kt
@@ -1,0 +1,12 @@
+package com.datadog.gradle.plugin.internal.sanitizer
+
+internal class ShortUriGitRemoteUrlSanitizer : FullUriGitRemoteUrlSanitizer() {
+    override fun sanitize(url: String): String {
+        val sanitizedUri = super.sanitize("$SSH_SCHEMA_PREFIX$url")
+        return sanitizedUri.removePrefix(SSH_SCHEMA_PREFIX)
+    }
+
+    companion object {
+        const val SSH_SCHEMA_PREFIX = "ssh://"
+    }
+}
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/UriParsingException.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/UriParsingException.kt
@@ -1,0 +1,6 @@
+package com.datadog.gradle.plugin.internal.sanitizer
+
+import java.lang.IllegalArgumentException
+
+internal class UriParsingException(message: String, throwable: Throwable) :
+    IllegalArgumentException(message, throwable)
--- a/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
+++ b/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
@@ -2,6 +2,8 @@ package com.datadog.gradle.plugin.internal
 
 import com.datadog.gradle.plugin.Configurator
 import com.datadog.gradle.plugin.RepositoryDetector
+import com.datadog.gradle.plugin.internal.sanitizer.UrlSanitizer
+import com.nhaarman.mockitokotlin2.whenever
 import fr.xgouchet.elmyr.Forge
 import fr.xgouchet.elmyr.annotation.StringForgery
 import fr.xgouchet.elmyr.junit5.ForgeConfiguration
@@ -18,6 +20,7 @@ import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.extension.Extensions
 import org.junit.jupiter.api.io.TempDir
+import org.mockito.Mock
 import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.junit.jupiter.MockitoSettings
 import org.mockito.quality.Strictness
@@ -35,14 +38,20 @@ internal class GitRepositoryDetectorTest {
     @TempDir
     lateinit var tempDir: File
 
-    @StringForgery(regex = "git@github\\.com:[a-z]+/[a-z][a-z0-9_-]+\\.git")
+    @StringForgery(regex = "http[s]?://github\\.com:[1-9]{2}/[a-z]+/repository\\.git")
     lateinit var fakeRemoteUrl: String
+
+    @StringForgery(regex = "http[s]?://github\\.com:[1-9]{2}/[a-z]+/repository\\.git")
+    lateinit var fakeSanitizedUrl: String
 
     lateinit var fakeProject: Project
 
     lateinit var fakeSourceSetFolders: List<File>
 
     lateinit var fakeTrackedFiles: List<String>
+
+    @Mock
+    lateinit var mockedUrlSanitizer: UrlSanitizer
 
     @Suppress("UnstableApiUsage")
     @BeforeEach
@@ -52,9 +61,10 @@ internal class GitRepositoryDetectorTest {
             .build()
 
         initializeSourceSets(forge)
-
+        whenever(mockedUrlSanitizer.sanitize(fakeRemoteUrl)).thenReturn(fakeSanitizedUrl)
         testedDetector = GitRepositoryDetector(
-            DefaultExecOperations((fakeProject as DefaultProject).processOperations)
+            DefaultExecOperations((fakeProject as DefaultProject).processOperations),
+            mockedUrlSanitizer
         )
     }
 
@@ -69,7 +79,7 @@ internal class GitRepositoryDetectorTest {
         // Then
         assertThat(result).hasSize(1)
         val repository = result.first()
-        assertThat(repository.url).isEqualTo(fakeRemoteUrl)
+        assertThat(repository.url).isEqualTo(fakeSanitizedUrl)
         assertThat(repository.hash).isNotNull().isNotBlank()
         assertThat(repository.sourceFiles)
             .containsExactlyInAnyOrder(*fakeTrackedFiles.toTypedArray())
@@ -104,7 +114,7 @@ internal class GitRepositoryDetectorTest {
         fakeTrackedFiles = trackedFiles
     }
 
-    private fun initializeGit() {
+    private fun initializeGit(remoteUrl: String = fakeRemoteUrl) {
         val readme = File(tempDir, "README.md")
         readme.writeText("# Fake project")
 
@@ -139,7 +149,7 @@ internal class GitRepositoryDetectorTest {
                 .waitForSuccess(5, TimeUnit.SECONDS)
         )
         check(
-            ProcessBuilder("git", "remote", "add", "origin", fakeRemoteUrl)
+            ProcessBuilder("git", "remote", "add", "origin", remoteUrl)
                 .directory(tempDir)
                 .start()
                 .waitForSuccess(5, TimeUnit.SECONDS)
--- a/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/sanitizer/GitRemoteUrlSanitizerTest.kt
+++ b/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/sanitizer/GitRemoteUrlSanitizerTest.kt
@@ -1,0 +1,438 @@
+package com.datadog.gradle.plugin.internal.sanitizer
+
+import com.datadog.gradle.plugin.Configurator
+import com.nhaarman.mockitokotlin2.doThrow
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.whenever
+import fr.xgouchet.elmyr.annotation.IntForgery
+import fr.xgouchet.elmyr.annotation.StringForgery
+import fr.xgouchet.elmyr.junit5.ForgeConfiguration
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import java.lang.RuntimeException
+import java.util.Locale
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.catchThrowable
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+internal class GitRemoteUrlSanitizerTest {
+    lateinit var testedUrlSanitizer: GitRemoteUrlSanitizer
+
+    @BeforeEach
+    fun `set up`() {
+        testedUrlSanitizer = GitRemoteUrlSanitizer()
+    }
+
+    // region Tests
+
+    @Test
+    fun `M sanitize a HTTP url W sanitize()`(
+        @StringForgery(regex = "http[s]?")
+        fakeSchema: String,
+        @StringForgery(
+            regex = "[a-z0-9]+\\.[a-z]{3}"
+        )
+        fakeHost: String,
+        @IntForgery(1, 99999) fakePort: Int,
+        @StringForgery fakeOwner: String,
+        @StringForgery fakeRepository: String
+    ) {
+        // Given
+        val fakeRemoteUrl = generateFakeFullRemoteUrl(
+            fakeSchema,
+            fakePort,
+            null,
+            null,
+            fakeHost,
+            fakeOwner,
+            fakeRepository
+        )
+
+        // Then
+        assertThat(testedUrlSanitizer.sanitize(fakeRemoteUrl)).isEqualTo(fakeRemoteUrl)
+    }
+
+    @Test
+    fun `M sanitize a HTTP url W sanitize() { username and password provided }`(
+        @StringForgery(regex = "http[s]?")
+        fakeSchema: String,
+        @StringForgery fakeUserName: String,
+        @StringForgery fakePassword: String,
+        @StringForgery(
+            regex = "[a-z0-9]+\\.[a-z]{3}"
+        )
+        fakeHost: String,
+        @IntForgery(1, 99999) fakePort: Int,
+        @StringForgery fakeOwner: String,
+        @StringForgery fakeRepository: String
+    ) {
+        // Given
+        val fakeRemoteUrl = generateFakeFullRemoteUrl(
+            fakeSchema,
+            fakePort,
+            fakeUserName,
+            fakePassword,
+            fakeHost,
+            fakeOwner,
+            fakeRepository
+        )
+
+        // Then
+        assertThat(testedUrlSanitizer.sanitize(fakeRemoteUrl)).isEqualTo(
+            generateFakeFullRemoteUrl(
+                fakeSchema,
+                fakePort,
+                null,
+                null,
+                fakeHost,
+                fakeOwner,
+                fakeRepository
+            )
+        )
+    }
+
+    @Test
+    fun `M sanitize a HTTP url W sanitize() { username provided }`(
+        @StringForgery(regex = "http[s]?")
+        fakeSchema: String,
+        @StringForgery fakeUserName: String,
+        @StringForgery(
+            regex = "[a-z0-9]+\\.[a-z]{3}"
+        )
+        fakeHost: String,
+        @IntForgery(1, 99999) fakePort: Int,
+        @StringForgery fakeOwner: String,
+        @StringForgery fakeRepository: String
+    ) {
+        // Given
+        val fakeRemoteUrl = generateFakeFullRemoteUrl(
+            fakeSchema,
+            fakePort,
+            fakeUserName,
+            null,
+            fakeHost,
+            fakeOwner,
+            fakeRepository
+        )
+
+        // Then
+        assertThat(testedUrlSanitizer.sanitize(fakeRemoteUrl)).isEqualTo(
+            generateFakeFullRemoteUrl(
+                fakeSchema,
+                fakePort,
+                null,
+                null,
+                fakeHost,
+                fakeOwner,
+                fakeRepository
+            )
+        )
+    }
+
+    @Test
+    fun `M sanitize a short format SSH url W sanitize()`(
+        @StringForgery fakeUserName: String,
+        @StringForgery(
+            regex = "[a-z0-9]+\\.[a-z]{3}"
+        )
+        fakeHost: String,
+        @StringForgery fakeOwner: String,
+        @StringForgery fakeRepository: String
+    ) {
+        // Given
+        val fakeRemoteUrl = generateFakeShortFormatRemoteUrl(
+            fakeUserName,
+            fakeHost,
+            fakeOwner,
+            fakeRepository
+        )
+
+        // Then
+        assertThat(testedUrlSanitizer.sanitize(fakeRemoteUrl)).isEqualTo(fakeRemoteUrl)
+    }
+
+    @Test
+    fun `M sanitize a SSH url W sanitize()`(
+        @StringForgery(
+            regex = "[a-z0-9]+\\.[a-z]{3}"
+        )
+        fakeHost: String,
+        @IntForgery(1, 99999) fakePort: Int,
+        @StringForgery fakeOwner: String,
+        @StringForgery fakeRepository: String
+    ) {
+        // Given
+        val fakeRemoteUrl = generateFakeFullRemoteUrl(
+            "ssh",
+            fakePort,
+            null,
+            null,
+            fakeHost,
+            fakeOwner,
+            fakeRepository
+        )
+
+        // Then
+        assertThat(testedUrlSanitizer.sanitize(fakeRemoteUrl)).isEqualTo(fakeRemoteUrl)
+    }
+
+    @Test
+    fun `M sanitize a SSH url W sanitize() { username and password provided }`(
+        @StringForgery fakeUserName: String,
+        @StringForgery fakePassword: String,
+        @StringForgery(
+            regex = "[a-z0-9]+\\.[a-z]{3}"
+        )
+        fakeHost: String,
+        @IntForgery(1, 99999) fakePort: Int,
+        @StringForgery fakeOwner: String,
+        @StringForgery fakeRepository: String
+    ) {
+        // Given
+        val fakeRemoteUrl = generateFakeFullRemoteUrl(
+            "ssh",
+            fakePort,
+            fakeUserName,
+            fakePassword,
+            fakeHost,
+            fakeOwner,
+            fakeRepository
+        )
+
+        // Then
+        assertThat(testedUrlSanitizer.sanitize(fakeRemoteUrl)).isEqualTo(
+            generateFakeFullRemoteUrl(
+                "ssh",
+                fakePort,
+                null,
+                null,
+                fakeHost,
+                fakeOwner,
+                fakeRepository
+            )
+        )
+    }
+
+    @Test
+    fun `M sanitize a SSH url W sanitize() { username provided }`(
+        @StringForgery fakeUserName: String,
+        @StringForgery(
+            regex = "[a-z0-9]+\\.[a-z]{3}"
+        )
+        fakeHost: String,
+        @IntForgery(1, 99999) fakePort: Int,
+        @StringForgery fakeOwner: String,
+        @StringForgery fakeRepository: String
+    ) {
+        // Given
+        val fakeRemoteUrl = generateFakeFullRemoteUrl(
+            "ssh",
+            fakePort,
+            fakeUserName,
+            null,
+            fakeHost,
+            fakeOwner,
+            fakeRepository
+        )
+
+        // Then
+        assertThat(testedUrlSanitizer.sanitize(fakeRemoteUrl)).isEqualTo(
+            generateFakeFullRemoteUrl(
+                "ssh",
+                fakePort,
+                null,
+                null,
+                fakeHost,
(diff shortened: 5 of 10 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -40,6 +42,19 @@ internal class GitRepositoryDetector : RepositoryDetector {
 
     // region Internal
 
+    private fun sanitizeUrl(remoteUrl: String): String {
+        return try {
+            val parsedUrl = URL(remoteUrl)
+            if (parsedUrl.userInfo != null) {
```
0xnm, line 48: just an open thought: I'm wondering if this covers all the cases of passing credentials in the remote URL. It seems, by looking on https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols that it should. Also [RFC2396](https://www.ietf.org/rfc/rfc2396.txt) specifies user info indeed. The funny thing, that there is a statement in RFC:

```
Some URL schemes use the format "user:password" in the userinfo
   field. This practice is NOT RECOMMENDED, because the passing of
   authentication information in clear text (such as URI) has proven to
   be a security risk in almost every case where it has been used.
```

Alternatively we could use `remoteUrl` field of plugin configuration.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
```diff
@@ -70,6 +70,83 @@ internal class GitRepositoryDetectorTest {
...
+        // Given
+        initializeGit(fakeRemoteUrlWithCredentials)
+
+        // When
+        val result = testedDetector.detectRepositories(fakeProject, fakeSourceSetFolders)
+
+        // Then
+        assertThat(result).hasSize(1)
+        val repository = result.first()
+        assertThat(repository.url).matches("https://@github.com/(.+)/repository.git")
```
0xnm, line 87: shouldn't we drop `@` symbol as well after sanitization? is it a correct link which will be working properly with that symbol left?

mariusc83: actually I am going to change the way I am sanitizing that URL as I don't like it now. I will just rebuild the URL by removing the `userInformation`

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -40,6 +42,24 @@ internal class GitRepositoryDetector : RepositoryDetector {
 
     // region Internal
 
+    private fun sanitizeUrl(remoteUrl: String): String {
+        return try {
+            val parsedUri = URI(remoteUrl)
+            val sanitizedUri = URI(
```
mariusc83, line 48: I think we should be ok now and this need to cover all the url formats.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
mariusc83: I changed the way we do the sanitization. I think we should be safe now but yeah adding an additional field for specifying the `remoteUrl` in the plugin configuration could be another idea.

xgouchet (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -39,6 +41,24 @@ internal class GitRepositoryDetector : RepositoryDetector {
...
+                null,
+                parsedUri.getHost(),
+                parsedUri.port,
+                parsedUri.getPath(),
+                parsedUri.getQuery(),
+                parsedUri.fragment
+            )
+            sanitizedUri.toURL().toString()
+        } catch (e: URISyntaxException) {
+            remoteUrl
```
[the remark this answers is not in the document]
mariusc83: Do you think we should ? What if the `remoteUrl` is of SSH format: `git@github.com:DataDog/dd-sdk-android-gradle-plugin.git`

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -39,6 +41,24 @@ internal class GitRepositoryDetector : RepositoryDetector {
     }
 
     // region Internal
+    @Suppress("SwallowedException")
+    private fun sanitizeUrl(remoteUrl: String): String {
+        return try {
+            val parsedUri = URI(remoteUrl)
```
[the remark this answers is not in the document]
mariusc83: hmm...that might work, it will at least skip one step of unnecessary logic.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -39,6 +41,24 @@ internal class GitRepositoryDetector : RepositoryDetector {
...
+            val sanitizedUri = URI(
+                parsedUri.scheme,
+                null,
+                parsedUri.getHost(),
+                parsedUri.port,
+                parsedUri.getPath(),
+                parsedUri.getQuery(),
+                parsedUri.fragment
+            )
+            sanitizedUri.toURL().toString()
```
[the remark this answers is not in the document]
mariusc83: Not sure....I just thought the `toString` method will not format it correctly, will try without `toURL`

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -39,6 +41,32 @@ internal class GitRepositoryDetector : RepositoryDetector {
...
+                    parsedUri.port,
+                    parsedUri.path,
+                    parsedUri.query,
+                    parsedUri.fragment
+                )
+                sanitizedUri.toString()
+            }
+        } catch (e: URISyntaxException) {
+            LOGGER.warn(
+                "The detected GIT remote url: [$remoteUrl] is not of type SSL." +
```
0xnm, line 64: the interesting point is that per https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols SSH remote URL can be one of the 2 formats:

* `ssh://[user@]server/project.git`
* or SCP-like syntax `[user@]server:project.git` (I missed it when first checked the page)

So former case should be handled fine by the branch above (with `URI` parsing), but latter case won't be and it still should be sanitized - I'm not expert in SCP remote format, but even if it allows just to specify only user name without a password, it is still a sensitive info.

0xnm (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
mariusc83: Oh...that's interesting, was actually wondering if the user information could be provided through SSH either. In this case I am not sure how I could avoid using `RegEx` to strip that part...not sure if I could find an SSH parser in Java, will have a look. Problem with `RegEx` is that it might not work if the format is not exactly the one you expect it to be and the question is : if we use `RegEx` for SSH then maybe we should have a rule that can be applied for both SSL or SSH formats and not use the `URI` parser anymore. What do you both think ? @xgouchet @0xnm

0xnm: I would say that we need only browsable remote url, which is only `http[s]://` schema and not `ssh://`, `git://` or whatever it is, even schema-less like in case of SCP syntax. Currently we send whatever is the output of the `git remote get-url origin` command. But maybe we can simplify it? Like don't send anything what doesn't start with `http[s]://` and if it should be sent, then sanitize it (which limits it to the URLs starting with `http[s]://` then).

As I understand, the point of this feature is to just have a browsable link, don't think we are interested in non-HTTP URLs.

mariusc83: Yes...reason why I keep it like that is that in the JS parser they actually have a `try/catch` mechanism and in case the url is not of `http[s]://(.+)` format they are just returning the unparsed URL. But as @xgouchet stated also there I think we should not allow at all non `http[s]` formats and we can even fail the build here with an error message.

mariusc83: I will update the PR accordingly

0xnm: Not sure we should fail a build. I'm using `git@github.com:DataDog/dd-sdk-android.git` for `dd-sdk-android` and it is perfectly valid URL, I would be a bit disappointed if something in the build doesn't like it. Maybe it may give just a warning instead.

If we fail the build, it may be CI build, which will add a lot of burden for the people watching it.

0xnm: Also it may be a company policy to use SSH-like URLs with cert auth instead of HTTPS, so they cannot change it.

mariusc83: Yes...I think I will just send a warning message, also please note that there is another PR (draft for now) which I just opened where we are letting the client to customize the `gitRemoteUrl` through the plugin extension. This will solve the case when you are using `SSH` . In conclusion using a warn message here will make the user aware that he needs to set an HTTP[S] url format manually to have the plugin working correctly.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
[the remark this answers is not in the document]
xgouchet: In case of a doubt, you have the source code of the URI class available ;)

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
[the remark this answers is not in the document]
xgouchet: Well then maybe we can detect this special case and treat it differently, but we should never swallow an exception silently

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
xgouchet: We need to be able to support standard remote url formats, and SSH urls `git@…` **are** a standard (it's used at least by Github, Gitlab, Bitbucket, …). The end goal is to be able to create a browsable link from a crash report but the backend, AFAIK, can convert the URIs in standard formats into a url (e.g.: `git@github.com:user/repository.git` → `https://github.com/user/repository/`).

mariusc83: Oh...this changes the approach then, so in this case we will keep that URL and probably sanitize it differently (through a RegEx maybe) but still send it to the backend after. Do we agree on this @xgouchet @0xnm

xgouchet: One option to do this without having to write a url compliant regex (which is going to be complex and tedious to write and test) is to search for the `"://"` substring to check wether or not the URI has a scheme. If not, you can prepend the string with `"ssh://"` and it will be parsed accordingly.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -74,4 +100,11 @@ internal class GitRepositoryDetector : RepositoryDetector {
     }
 
     // endregion
+
+    companion object {
+        const val WRONG_URL_FORMAT_WARNING_MESSAGE =
+            "The detected GIT remote url: [%s] is not of 'http[s]' and it will" +
+                "not work properly with this plugin. You can provide a different URL through" +
```
0xnm, line 107:
```suggestion
                " not work properly with this plugin. You can provide a different URL through" +
```

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -74,4 +100,11 @@ internal class GitRepositoryDetector : RepositoryDetector {
...
+                "the plugin extension."
```
0xnm, line 108:
```suggestion
                " the plugin extension."
```

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -39,6 +42,29 @@ internal class GitRepositoryDetector : RepositoryDetector {
     }
 
     // region Internal
+    @Suppress("SwallowedException")
+    private fun sanitizeUrl(remoteUrl: String): String {
+        return try {
+            val parsedUri = URI(remoteUrl)
+            if (parsedUri.userInfo.isNullOrEmpty()) {
+                parsedUri.toString()
```
0xnm, line 50: very minor: we can also return original string here

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
mariusc83: Hmmm the last option seems more reliable and I can apply the same logic after by using the URI. I will try to go with this approach.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
0xnm, line 107: just a thought: user may still wonder what will not work properly, maybe we can just say it won't be browsable from dashboard? or make a link to the feature. wdyt?

0xnm (approved):
lgtm! just few minor comments/suggestions

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -39,6 +42,29 @@ internal class GitRepositoryDetector : RepositoryDetector {
     }
 
     // region Internal
+    @Suppress("SwallowedException")
```
0xnm, line 45: is this still needed given that exception is logged?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
0xnm: I wasn't aware about this backend feature, that it can do the conversion. The interesting part is that if remote is of SCP syntax, like `ssh://user@server:project.git`, then adding `ssh://` helps to parse it, but `getUserInfo` returns `null` https://ideone.com/eUkYAV

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/DdUrlSanitizer.kt
```diff
@@ -0,0 +1,41 @@
...
+ * It will strip out the user information from the provided URL by applying the following rules:
+ * 1. In case the URL is a valid schema URL : ((ssh|http[s]?):\\) it will strip out the user
+ * information if any from it
+ * 2. In case the URL is a valid schema URL but by mistake the user forgot to add the schema:
+ * [username]:[password]@github.com/(.+).git it will strip out the user information if any and will
+ * return the new url
+ * 3. In case the URL is of a GIT SSH short format: [user@]github.com:(.+).git it will do
+ * nothing assuming that there is no password normally provided in a GIT SSH short format URL.
+ */
+internal class DdUrlSanitizer(
```
xgouchet, line 15: This name is not really relevant: the `Dd` prefix doesn't provide information, and is useless in the sense that all the code base is in the `dd-sdk-android-gradle-plugin` already. 

If we assume that the `UrlSanitizer` interface aims to "cleanup a url to be usable in DD", then this implementation specific focus is to "strip out the user information from the provided URL" as mentioned above. I'd name this `UserInfoUrlSanitizer` or something similar.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -3,14 +3,20 @@ package com.datadog.gradle.plugin.internal
...
 import java.io.File
 import org.gradle.api.Project
 import org.gradle.process.internal.ExecException
 
 // TODO RUMM-1095 handle git submodules
 // TODO RUMM-1096 handle git subtrees
 // TODO RUMM-1093 let customer override `origin` with custom remote name
-internal class GitRepositoryDetector : RepositoryDetector {
+internal class GitRepositoryDetector(
+    private val urlSanitizerFactory: () -> UrlSanitizer = {
```
xgouchet, line 16: Why are you providing a factory and not an instance directly ?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/SanitizerStrategy.kt
```diff
@@ -0,0 +1,5 @@
+package com.datadog.gradle.plugin.internal.sanitizer
+
+internal interface SanitizerStrategy {
```
xgouchet, line 3: Why do you need this specific interface. It has the same purpose and signature as `UrlSanitizer`.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/UriParsingException.kt
```diff
@@ -0,0 +1,6 @@
+package com.datadog.gradle.plugin.internal.sanitizer
+
+import java.rmi.RemoteException
+
+internal class UriParsingException(message: String, throwable: Throwable) :
```
xgouchet, line 5: I'm not sure a `RemoteException` is the appropriate parent class for this (especially coming from the `java.rmi` package : 
>  A `RemoteException` is the common superclass for a number of communication-related exceptions that may occur during the execution of a remote method call

If we do need a custom exception, I'd make it a child of either `java.net.MalformedURLException` or `java.lang.IllegalArgumentException`.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/DefaultSanitizerStrategy.kt
```diff
@@ -0,0 +1,26 @@
+package com.datadog.gradle.plugin.internal.sanitizer
+
+import java.net.URI
+import java.net.URISyntaxException
+import kotlin.jvm.Throws
+
+internal class DefaultSanitizerStrategy : SanitizerStrategy {
```
xgouchet, line 7: `Default` is not a meaningful prefix. I'd name that `FullUriUserInfoSanitizer`

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/NoSchemaSanitizerStrategy.kt
```diff
@@ -0,0 +1,27 @@
+package com.datadog.gradle.plugin.internal.sanitizer
+
+import java.net.URI
+
+internal class NoSchemaSanitizerStrategy : SanitizerStrategy {
```
xgouchet, line 5: Out of consistency I'd name that `ShortUriUserInfoSanitizer`

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/DdUrlSanitizer.kt
```diff
@@ -0,0 +1,41 @@
...
+            return strategyResolver(url).apply(url)
+        } catch (e: Exception) {
+            throw UriParsingException(WRONG_URL_FORMAT_ERROR_MESSAGE.format(url, Locale.US), e)
+        }
+    }
+
+    companion object {
+        const val WRONG_URL_FORMAT_ERROR_MESSAGE =
+            "The detected GIT remote url: [%s] is not a valid GIT url and it " +
+                "will not be browsable when using it from error dashboard. " +
```
xgouchet, line 37:
```suggestion
                "will not be browsable from the Error Tracking panel. " +
```

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/DdUrlSanitizer.kt
```diff
@@ -0,0 +1,41 @@
...
+                "You can provide a different URL through the plugin extension."
+        const val VALID_SCHEMA_URL_FORMAT_REG_EX = "^(ssh|http|https)://(.+)$"
```
xgouchet, line 39: I'd rename that `…_PATTERN` and create a static `…_REGEX = Regex(VALID_SCHEMA_URL_FORMAT_PATTERN)`

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/NoSchemaSanitizerStrategy.kt
```diff
@@ -0,0 +1,27 @@
...
+    override fun apply(url: String): String {
```
xgouchet, line 6: This one could be simplified by doing 
```kotlin
internal class ShortUriUserInfoSanitizer : FullUriUserInfoSanitizer {
    override fun apply(url: String): String {
        return super.apply("$SSH_SCHEMA_PREFIX$url")
                    .removePrefix(SSH_SCHEMA_PREFIX)
    }
}
```

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/sanitizer/DdUrlSanitizerTest.kt
```diff
@@ -0,0 +1,157 @@
...
+
+    @Test
+    fun `M sanitize a HTTP url W sanitize() { username and password provided }`(
+        @StringForgery(
+            regex = "http[s]?://[a-z]+:[a-z]+@github\\.com:[1-9]{2}/[a-z]+/[a-z0-9_]+\\.git"
+        )
+        fakeRemoteUrl: String
+    ) {
+        assertThat(testedUrlSanitizer.sanitize(fakeRemoteUrl))
+            .matches("http[s]?://github.com:[1-9]{2}/[a-z]+/[a-z0-9_]+\\.git")
```
xgouchet, line 52: Technically this assertion is not enough. If the sanitizer changed the port number or the path segments, it would still pass but it won't be correct. 
We can instead do something like: 
```kotlin
    fun `M sanitize a HTTP url W sanitize() { username and password provided }`(
        @StringForgery fakeUserName: String,
        @StringForgery fakePassword: String,
        @StringForgery(regex = "[a-z0-9]+\\.[a-z]{3}" fakeHost: String,
        @IntForgery(1, 99999) fakePort: Int,
        @StringForgery fakeOwner: String,
        @StringForgery fakeRepository: String
    ) {
        // Given
        val fakeRemoteUrl = "http://$fakeUserName:$fakePassword@$fakeHost/$fakeOwner/$fakeRepository.git"

        // When
        val result = testedUrlSanitizer.sanitize(fakeRemoteUrl)

        // Then
        assertThat(result)
            .isEqualTo("http://$fakeHost/$fakeOwner/$fakeRepository.git")
```

And the same comment applies to all the test functions in this class

xgouchet (changes requested):

0xnm (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/UriParsingException.kt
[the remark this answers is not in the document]
mariusc83: oh...my bad my intention was to use `RuntimeException` is just the IDE autocomplete playing tricks on me.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/DdUrlSanitizer.kt
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+) : UrlSanitizer {
+
+    @SuppressWarnings("TooGenericExceptionCaught")
+    override fun sanitize(url: String): String {
+        try {
+            return strategyResolver(url).apply(url)
+        } catch (e: Exception) {
+            throw UriParsingException(WRONG_URL_FORMAT_ERROR_MESSAGE.format(url, Locale.US), e)
```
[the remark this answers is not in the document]
mariusc83: This is only to make sure we are not missing anything, normally this should never occur as this can only be triggered from the `URI` parsing and honestly I tried with all the combinations and as long as you have the schema set there is no way to trigger an exception from there :). We are already checking for a schema existence before going into the `URI` parser but just for the code sanity we need to make sure we are treating that possible exception. Unfortunately if there is any error in the `URI` parser the plugin will be broken as that `URL` will not be usable and we will have to fail the build.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/DdUrlSanitizer.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Locale
+
+/**
+ * It will strip out the user information from the provided URL by applying the following rules:
+ * 1. In case the URL is a valid schema URL : ((ssh|http[s]?):\\) it will strip out the user
+ * information if any from it
+ * 2. In case the URL is a valid schema URL but by mistake the user forgot to add the schema:
+ * [username]:[password]@github.com/(.+).git it will strip out the user information if any and will
+ * return the new url
+ * 3. In case the URL is of a GIT SSH short format: [user@]github.com:(.+).git it will do
```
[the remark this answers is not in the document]
mariusc83: I agree but in the same time after a discussion with @xgouchet we came to the conclusion that leaking the `username` will not be that bad in some situations, otherwise we will have to build a complex RegEx mechanism to threat this case.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/UriParsingException.kt
mariusc83: Yes...as I explained to @0xnm this was a mistake, my intention was to use a `RuntimeException` is just the editor playing tricks on me.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
mariusc83: Good point...I needed this in a previous implementation but yes, not needed anymore

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/SanitizerStrategy.kt
mariusc83: I'd like to keep things separated in case we want to go back to this later. This feature might need improvements later or new strategies to handle different other cases so it will be easily scalable this way. It is also better for testing as I can handle the `Strategy` output better.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/sanitizer/DdUrlSanitizerTest.kt
mariusc83: Yes...I do agree, I think it's better to provide separated values for all this parameters.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/SanitizerStrategy.kt
mariusc83: Nevermind...I see what you're saying...will remove that and make this as a wrapper of `UrlSantizers` implementations.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/NoSchemaSanitizerStrategy.kt
mariusc83: In the same time this will only work with one specific prefix : `ssh:\\` in this case, I'd rather keep it like this which applies for all prefixes and it is also one operation in minus having not to do that `search and replace` in the String at the end.

0xnm (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/FullUriGetRemoteUrlSanitizer.kt
```diff
@@ -0,0 +1,26 @@
+package com.datadog.gradle.plugin.internal.sanitizer
+
+import java.net.URI
+import java.net.URISyntaxException
+import kotlin.jvm.Throws
+
+internal class FullUriGetRemoteUrlSanitizer : UrlSanitizer {
```
xgouchet, line 7: There's a typo, I think you mean `FullUriGitRemoteUrlSanitizer`

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/sanitizer/ShortUriGitRemoteUrlSanitizer.kt
```diff
@@ -0,0 +1,27 @@
+package com.datadog.gradle.plugin.internal.sanitizer
+
+import java.net.URI
+
+internal class ShortUriGitRemoteUrlSanitizer : UrlSanitizer {
```
xgouchet, line 5: It might be a good idea to let this extend the `FullUriGitRemoteUrlSanitizer`. If at some point we need to change the sanitization rule, you could reuse the same logic and only have the change done in one place: 

```
override fun sanitize(url: String): String {
    val sshUri ="$SSH_SCHEMA_PREFIX$url"
    val sanitizedUri = super.sanitize(sshUri)
    return sanitizedUri.removePrefix(SSH_SCHEMA_PREFIX)
}
```

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetectorTest.kt
```diff
@@ -33,24 +36,30 @@ internal class GitRepositoryDetectorTest {
...
     @BeforeEach
     fun `set up`(forge: Forge) {
         fakeProject = ProjectBuilder.builder()
             .withProjectDir(tempDir)
             .build()
 
         initializeSourceSets(forge)
-
-        testedDetector = GitRepositoryDetector()
+        whenever(mockedUrlSanitizer.sanitize(fakeRemoteUrl)).thenReturn(fakeSanitizedUrl)
```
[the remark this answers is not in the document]
xgouchet: This is to ensure that the url we report is always the sanitized one, whatever the sanitization result is. The UrlSanitizer being a mock we don't want to make assumption on what rules it uses.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/sanitizer/GitRemoteUrlSanitizerTest.kt
```diff
@@ -0,0 +1,439 @@
...
+        } else if (password != null) {
+            "$schema://$password@$host:$port/$owner/$repository.git"
+        } else if (username != null) {
+            "$schema://$username@$host:$port/$owner/$repository.git"
+        } else {
+            "$schema://$host:$port/$owner/$repository.git"
+        }
+    }
+
+    private fun generateFakeMissingSchemaRemoteUrl(
```
xgouchet, line 411: You might combine both those methods in one: 
```kotlin
    private fun generateFakeFullRemoteUrl(
        schema: String?,
        port: Int,
        username: String?,
        password: String?,
        host: String,
        owner: String,
        repository: String
    ): String {
        val builder = StringBuilder()
        if (schema != null) {
            builder.append(schema)
            builder.append("://")
        }
        if (username != null) {
            builder.append(username)
            if (password != null) {
                builder.append(":")
                builder.append(username)
            }
            builder.append("@")
        }
        builder.append(host)
        builder.append(":")
        builder.append(port)
        builder.append("/")
        builder.append(owner)
        builder.append("/")
        builder.append(repository)
        builder.append(".git")
    }
```

xgouchet (changes requested):

xgouchet (approved):

0xnm (approved):

PR #47, opened 2021-09-03
Add some logs to diagnose issues

xgouchet:


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -72,12 +73,12 @@ class DdAndroidGradlePlugin @Inject constructor(
         extension: DdExtension
     ): Task? {
         if (!variant.buildType.isMinifyEnabled) {
-            // Proguard/R8 are not enabled, let's not create a task
+            LOGGER.warn("Minifying disabled for variant ${variant.name}, no upload task created")
```
0xnm, line 76: I would probably use `info` level instead of `warning` here and below to not confuse users, because there is nothing wrong with having variants where minification is not enabled or extension is disabled. Also warning will be shown in the console by default, but `info` requires having `--info` flag during task run, which is better suits debugging ideology.

But anyway, this is minor, I'm fine with warning as well.

0xnm (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
xgouchet: I know, but the thing is that most users won't think about using the `--info` if something goes wrong. If the warning is there it might be noisy, but it means they'll see it and it might avoid some unnecessary calls to support. WDYT?

0xnm: Do we have requests about issues with the plugin very often? I guess experienced users can look up `TROUBLESHOOTING` doc (and we can add note about `--info` there) or already know about `--info` flag. Or our support can advise to do the same to speed up investigation process.

mariusc83 (approved):

Issue #45, opened 2021-08-02
Performance issue


NFesquet:
**Describe what happened**
When using Datadog plugin, Gradle build performance drops. I've been benchmarking with [Gradle Profiler](https://github.com/gradle/gradle-profiler), and there's an increase of build times by 6,2 on average.

See detailed report
![Screenshot 2021-08-02 at 10 20 31](https://user-images.githubusercontent.com/2885719/127828369-b77454e6-c88a-45fc-828d-97d438c6fe15.png)

First warm-up build without plugin took longer because Gradle Profiler ran for the 1st time. I should have ran it twice in a row and use 2nd report to prevent from impacting results.

**Steps to reproduce the issue:**
Add Datadog plugin to the project.

**Describe what you expected:**
Little to no increase of build times.

**Additional context**
 - Android Build Tools version: 30.0.2
 - Gradle version: 7.1.1
 - OS version: Mac OS Big Sur 11.4
 - Plugin version: 1.0.1
 - Proguard configuration: None (happens on debug builds)
 - Other Gradle Plugins: Kapt, Firebase Crashlytics, Firebase Performance, AppDynamics, EasyLauncher

xgouchet:
Hi @NFesquet, Can you give us more information about your actual project, to have a better clue on those numbers. Specifically, can you tell us: 
- What's the size of your proguard mapping files
- What's the number of source files (java/kotlin) in your project

NFesquet:
Hi @xgouchet! Sure, here are our stats:
The project has ~150 Java files, ~3200 Kotlin files and 49 Gradle modules.
Proguard mapping file has 152412 lines.

Hi @0xnm, I'll try that and get back to you asap.

NFesquet:
@0xnm I tried
 ```kotlin
datadog {
    checkProjectDependencies = com.datadog.gradle.plugin.SdkCheckLevel.NONE
}
```

But this option doesn't change build times much. If you want, I can run benchmarks to compare build times with and without the option, but I can already say that with it build times are still +40.000ms .

0xnm:
Hello @NFesquet. Can you please also tell us how many product flavors, flavor dimensions and build types do you have? Also if you are able to say which method of our plugin takes majority of the time by looking on the profiling details, that would be very helpful.

NFesquet:
[...]

PR #48, opened 2021-09-07
RUMM-1546 Introduce the CheckDepsTask

mariusc83:
### What does this PR do?

Following this reported [issue](https://github.com/DataDog/dd-sdk-android-gradle-plugin/issues/45) and after further investigations it seems that by doing:
```
variantCompileTasks.doLast { ... }
```
we were braking the `up-to-date` checks and therefore the incremental capability for the compilation tasks (the compile task was alway performed even if the code was not changed).
This PR proposes a new approach for the check dependency routine by introducing the new `CheckDepsTask` which will also take advantage of the task incremental capability (meaning that if the module configuration will not be changed this task will not be executed again).

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
@@ -9,14 +9,13 @@ package com.datadog.gradle.plugin
 import com.android.build.gradle.AppExtension
 import com.android.build.gradle.api.ApplicationVariant
 import com.datadog.gradle.plugin.internal.GitRepositoryDetector
-import com.datadog.gradle.plugin.internal.MissingSdkException
 import com.datadog.gradle.plugin.internal.VariantIterator
 import java.io.File
 import javax.inject.Inject
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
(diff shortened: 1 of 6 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -141,39 +139,21 @@ class DdAndroidGradlePlugin @Inject constructor(
...
+                extensionConfiguration.checkProjectDependencies ?: SdkCheckLevel.FAIL
+            val checkDepsTask = target.tasks.register(
+                checkDepsTaskName,
+                DdCheckSdkDepsTask::class.java
+            ) {
+                it.configuration.set(variant.compileConfiguration)
+                it.sdkCheckLevel.set(resolvedCheckDependencyFlag)
+                it.variantName.set(variant.name)
+            }.get()
+            compileTask.finalizedBy(checkDepsTask)
```
0xnm, line 155: maybe `finalizedBy` supports `TaskProvider` as argument? In that case we could avoid calling `get` above and it would be nice for the lazy graph evaluation. Theoretically it should by looking on this https://docs.gradle.org/current/dsl/org.gradle.api.Task.html#org.gradle.api.Task.dependencies (`TaskProvider` is children of `Provider`)

0xnm (approved):

xgouchet (approved):

PR #55, opened 2021-11-24
RUMM-1795 Ensure git files are reported with full path from repo root

xgouchet:
### What does this PR do?

Ensure git files are reported with full path from repo root.

### Motivation

In some cases the `cwd` is not the repo root, meaning that the path we get from `git ls-files` are not correct. Adding the `--full-name` forces git to report the paths from the repo root.

### DIFF
--- a/samples/basic/build.gradle
+++ b/samples/basic/build.gradle
@@ -34,7 +34,6 @@ dependencies {
     implementation(libs.androidx.constraintlayout)
 
     // implementation(libs.datadogSdk)
-
 }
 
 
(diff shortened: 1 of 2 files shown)


samples/basic/build.gradle
```diff
@@ -33,6 +33,7 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.2.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.0.4'
 
+    implementation 'com.datadoghq:dd-sdk-android:1.11.0-rc1'
```
0xnm, line 36: it should be rebased, `main` already has it (but commented) https://github.com/DataDog/dd-sdk-android-gradle-plugin/blob/4ebeb840a822b5c793fc1847f5cffcf819258f0b/samples/basic/build.gradle#L36

0xnm (changes requested):

mariusc83 (approved):

0xnm (approved):

PR #59, opened 2021-12-03
RUMM-1778 delay the android extension check

xgouchet:
### What does this PR do?

Delay the android extension check.

### Motivation

In some cases, the order in which plugins are applied doesn't match the order in the `plugins { }` block, meaning that the Datadog plugin can be started before the Android one is applied, meaning the extension is not there yet. Delaying it ensures the plugin gets applied properly

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
@@ -32,20 +32,19 @@ class DdAndroidGradlePlugin @Inject constructor(
     /** @inheritdoc */
     override fun apply(target: Project) {
 
-        val androidExtension = target.extensions.findByType(AppExtension::class.java)
-        if (androidExtension == null) {
-            LOGGER.error(ERROR_NOT_ANDROID)
-            return
-        }
-
         val extension = target.extensions.create(EXT_NAME, DdExtension::class.java)
         extension.variants = target.container(DdExtensionConfiguration::class.java)
         val apiKey = resolveApiKey(target)
 
         target.afterEvaluate {
-            androidExtension.applicationVariants.all { variant ->
-                configureVariantForUploadTask(target, variant, apiKey, extension)
-                configureVariantForSdkCheck(target, variant, extension)
+            val androidExtension = target.extensions.findByType(AppExtension::class.java)
+            if (androidExtension == null) {
+                LOGGER.error(ERROR_NOT_ANDROID)
+            } else {
+                androidExtension.applicationVariants.all { variant ->
+                    configureVariantForUploadTask(target, variant, apiKey, extension)
+                    configureVariantForSdkCheck(target, variant, extension)
+                }
             }
         }
     }


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -32,20 +32,19 @@ class DdAndroidGradlePlugin @Inject constructor(
...
-        if (androidExtension == null) {
-            LOGGER.error(ERROR_NOT_ANDROID)
-            return
-        }
-
         val extension = target.extensions.create(EXT_NAME, DdExtension::class.java)
         extension.variants = target.container(DdExtensionConfiguration::class.java)
         val apiKey = resolveApiKey(target)
 
         target.afterEvaluate {
```
0xnm, line 39: I was curious about `afterEvaluate` impact on the overall gradle caching mechanism and found the following thread https://discuss.gradle.org/t/is-project-afterevaluate-the-proper-way-for-gradle-plugin-to-dynamically-create-default-tasks/31349. However, `afterEvaluate` at the same time is used in the official Gradle docs to create tasks, so I guess we can still use it.

Also I checked what are the other options if we want to do the stuff after `Android` plugin is applied. It turns out that there is `project.plugins.withId` which can be executed only when given plugin is applied https://docs.gradle.org/current/javadoc/org/gradle/api/plugins/PluginContainer.html#withId-java.lang.String-org.gradle.api.Action-.

And also AGP has a special marker as class (if we won't want to use ID) https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-api/src/main/java/com/android/build/gradle/api/AndroidBasePlugin.kt;l=32;drc=ecdfaee5fbdfa69e82bb9266b6742d9c3db27880 which is applied when any AGP plugin is ready https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/plugins/BasePlugin.java;l=314;drc=5778dae21439b6353fbf10a73d1393afcafafc93.

This is another option compared to `afterEvaluate`. But I'm fine with `afterEvaluate` as well.

0xnm (approved):

mariusc83 (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -32,20 +32,19 @@ class DdAndroidGradlePlugin @Inject constructor(
...
-            androidExtension.applicationVariants.all { variant ->
-                configureVariantForUploadTask(target, variant, apiKey, extension)
-                configureVariantForSdkCheck(target, variant, extension)
+            val androidExtension = target.extensions.findByType(AppExtension::class.java)
```
mariusc83, line 40: but moving it into the `afterEvaluate` will not mean that you will create a `DdExtension` just for nothing into a non Android project ?

0xnm: true, this is a concern which I missed

xgouchet: It's not necessarily a concern in itself, the extension won't impact the project whatsoever.

Issue #62, opened 2021-12-07
Execution failed for task 'checkSdkDeps'


batadamnjanovic:
Hi all. 

We have a problem with applying Datadog Gradle plugin `com.datadoghq:dd-sdk-android-gradle-plugin:1.2.0`.

When the plugin is applied in the app's build.gradle script with `id("com.datadoghq.dd-sdk-android-gradle-plugin"), we get the following error when we run a build(assemble task).

We have two flavors: production and staging.
Both flavors have debug and release variants.
We have the app module that include several other modules as dependencies.

**Additional context**
 - Android Build Tools version: 4.1.3
 - Gradle version: 7.3.1
 - OS version: Mac OS Big Sur 11.4
 - Plugin version: 1.2.0

Stacktrace: 
`Execution failed for task ':app:checkSdkDepsStagingDebug'.
> Error while evaluating property 'configuration' of task ':app:checkSdkDepsStagingDebug'
   > Could not resolve all files for configuration ':app:stagingDebugCompileClasspath'.
      > The consumer was configured to find an API of a component, as well as attribute 'com.android.build.api.attributes.BuildTypeAttr' with value 'debug', attribute 'configuration' with value 'staging', attribute 'org.jetbrains.kotlin.platform.type' with value 'androidJvm'. However we cannot choose between the following variants of project :features:
          - Configuration ':features:stagingDebugApiElements' variant android-databinding-class-log declares an API of a component, as well as attribute 'com.android.build.api.attributes.BuildTypeAttr' with value 'debug', attribute 'configuration' with value 'staging', attribute 'org.jetbrains.kotlin.platform.type' with value 'androidJvm':
              - Unmatched attributes:
                  - Provides attribute 'artifactType' with value 'android-databinding-class-log' but the consumer didn't ask for it
                  - Provides attribute 'com.android.build.api.attributes.VariantAttr' with value 'stagingDebug' but the consumer didn't ask for it
          - Configuration ':features:stagingDebugApiElements' variant android-manifest declares an API of a component, as well as attribute 'com.android.build.api.attributes.BuildTypeAttr' with value 'debug', attribute 'configuration' with value 'staging', attribute 'org.jetbrains.kotlin.platform.type' with value 'androidJvm':
              - Unmatched attributes:
                  - Provides attribute 'artifactType' with value 'android-manifest' but the consumer didn't ask for it
[...]

PR #64, opened 2021-12-09
Create samples lib module to better test dependency check task

0xnm:
### What does this PR do?

This PR creates `sample:lib-module` module as a follow-up of #62 to be able to test multi-module setup.


samples/basic/build.gradle
```diff
@@ -27,6 +27,7 @@ android {
 }
 
 dependencies {
+    implementation(project(':samples:lib-module'))
```
xgouchet, line 30: Maybe in addition to that we could refer to the `Placeholder` class in the sample app code.

xgouchet (approved):

samples/basic/build.gradle
0xnm: I've checked final apk and `Placeholder` class is there (at least in `debug` variant) without referencing it explicitly from the code, but yes, let's add it in case if compiler becomes smarter.

PR #67, opened 2022-01-05
RUMM-1814 Provide functional tests for Datadog Gradle plugin

mariusc83:
### What does this PR do?

In this PR we are adding functional tests for the Datadog Gradle Plugin:
- testing `assemble` task with and without the `--build-cache` option
- testing `assemble` task with and without the `--configuration-cache` option
- testing `upload` task for all the flavours
- testing the plugin configuration options

### Motivation

### Additional Notes

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
@@ -114,7 +114,7 @@ fun Project.publishingConfig(projectDescription: String) {
         signingExtension.apply {
             val privateKey = System.getenv("GPG_PRIVATE_KEY")
             val password = System.getenv("GPG_PASSWORD")
-            isRequired = true
+            isRequired = !hasProperty("dd-skip-signing")
             useInMemoryPgpKeys(privateKey, password)
             sign(publishingExtension.publications.getByName(MavenConfig.PUBLICATION))
         }
--- a/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
+++ b/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
@@ -1,0 +1,583 @@
+package com.datadog.gradle.plugin
+
+import com.datadog.gradle.plugin.utils.initializeGit
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.annotation.StringForgery
+import fr.xgouchet.elmyr.junit5.ForgeConfiguration
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import java.io.File
+import java.util.Properties
+import org.assertj.core.api.Assertions.assertThat
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.gradle.testkit.runner.internal.PluginUnderTestMetadataReading
+import org.junit.Ignore
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import org.junit.jupiter.api.io.TempDir
+
+@Extensions(
+    ExtendWith(ForgeExtension::class)
+)
+@ForgeConfiguration(Configurator::class)
+internal class DdAndroidGradlePluginFunctionalTest {
+    @TempDir
+    lateinit var testProjectDir: File
+    lateinit var appRootDir: File
+    lateinit var libModuleRootDir: File
+    lateinit var appMainSrcDir: File
+    lateinit var appKotlinSourcesDir: File
+    lateinit var libModuleMainSrcDir: File
+    lateinit var libModuleKotlinSourcesDir: File
+    lateinit var settingsFile: File
+    lateinit var appBuildGradleFile: File
+    lateinit var libModuleBuildGradleFile: File
+    lateinit var appManifestFile: File
+    lateinit var libModuleManifestFile: File
+    lateinit var gradlePropertiesFile: File
+    lateinit var sampleApplicationClassFile: File
+    lateinit var libModulePlaceholderFile: File
+
+    @StringForgery(regex = "http[s]?://github\\.com:[1-9]{2}/[a-z]+/repository\\.git")
+    lateinit var fakeRemoteUrl: String
+    private val colors = listOf("Blue", "Green")
+    private val versions = listOf("Demo", "Full")
+    val variants: List<String> by lazy {
+        versions.flatMap { version ->
+            colors.map {
+                "${version.lowercase()}$it"
+            }
+        }
+    }
+
+    @BeforeEach
+    fun `set up`() {
+        appRootDir = File(testProjectDir, "samples/app").apply { mkdirs() }
+        libModuleRootDir = File(testProjectDir, "samples/lib-module").apply { mkdirs() }
+        settingsFile = File(testProjectDir, "settings.gradle")
+        gradlePropertiesFile = File(testProjectDir, "gradle.properties")
+        appMainSrcDir = File(appRootDir, "src/main").apply { mkdirs() }
+        appKotlinSourcesDir = File(appMainSrcDir, "kotlin").apply { mkdirs() }
+        libModuleMainSrcDir = File(libModuleRootDir, "src/main").apply { mkdirs() }
+        libModuleKotlinSourcesDir = File(libModuleMainSrcDir, "kotlin").apply { mkdirs() }
+        appBuildGradleFile = File(appRootDir, "build.gradle")
+        libModuleBuildGradleFile = File(libModuleRootDir, "build.gradle")
+        appManifestFile = File(appMainSrcDir, "AndroidManifest.xml")
+        libModuleManifestFile = File(libModuleMainSrcDir, "AndroidManifest.xml")
+        sampleApplicationClassFile = File(appKotlinSourcesDir, "SampleApplication.kt")
+        libModulePlaceholderFile = File(libModuleKotlinSourcesDir, "Placeholder.kt")
+
+        stubFile(settingsFile, SETTINGS_FILE_CONTENT)
+        stubFile(sampleApplicationClassFile, APPLICATION_CLASS_CONTENT)
+        stubFile(appManifestFile, APP_MANIFEST_FILE_CONTENT)
+        stubFile(gradlePropertiesFile, GRADLE_PROPERTIES_FILE_CONTENT)
+        stubFile(libModulePlaceholderFile, LIB_MODULE_PLACEHOLDER_CLASS_CONTENT)
+        stubFile(libModuleManifestFile, LIB_MODULE_MANIFEST_FILE_CONTENT)
+        stubGradleBuildFromResourceFile(
+            "lib_module_build.gradle.kts",
+            libModuleBuildGradleFile
+        )
+        initializeGit(fakeRemoteUrl, appRootDir)
+    }
+
+    // region Assemble
+
+    @Test
+    fun `M success W assembleRelease`() {
+        // Given
+        stubGradleBuildFromResourceFile(
+            "build_with_datadog_dep.gradle.kts",
+            appBuildGradleFile
+        )
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments(":samples:app:assembleRelease")
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":samples:app:assembleRelease")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @Test
+    fun `M success W assembleRelease { project with library module }`() {
+        // Given
+        stubGradleBuildFromResourceFile(
+            "build_with_lib_module_attached.gradle.kts",
+            appBuildGradleFile
+        )
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments(":samples:app:assembleRelease")
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":samples:app:assembleRelease")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @Test
+    fun `M success W assembleRelease { build cache }`() {
+        // Given
+        stubGradleBuildFromResourceFile(
+            "build_with_datadog_dep.gradle.kts",
+            appBuildGradleFile
+        )
+
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments("--build-cache", ":samples:app:assembleRelease")
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":samples:app:assembleRelease")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @Test
+    fun `M success W assembleDebug`() {
+        // Given
+        stubGradleBuildFromResourceFile(
+            "build_with_datadog_dep.gradle.kts",
+            appBuildGradleFile
+        )
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments(":samples:app:assembleDebug")
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":samples:app:assembleDebug")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @Test
+    fun `M success W assembleDebug { project with library module }`() {
+        // Given
+        stubGradleBuildFromResourceFile(
+            "build_with_lib_module_attached.gradle.kts",
+            appBuildGradleFile
+        )
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments(":samples:app:assembleDebug")
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":samples:app:assembleDebug")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @Test
+    fun `M success W assembleDebug { build cache }`() {
+        // Given
+        stubGradleBuildFromResourceFile(
+            "build_with_datadog_dep.gradle.kts",
+            appBuildGradleFile
+        )
+
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments("--build-cache", ":samples:app:assembleDebug")
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":samples:app:assembleDebug")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @Ignore(
+        "This test is ignored for now because of these tasks: " +
+            "collect[Flavour1][Flavour2]ReleaseDependencies. This is caused under the hood" +
+            "by this task: PerModuleReportDependenciesTask which accesses the project object" +
+            "inside the action method. " +
+            "There is already an opened issue here: https://github.com/gradle/gradle/issues/12871"
+    )
+    fun `M success W assembleRelease { configuration cache, checkProjectDependecies enabled }`() {
+        // TODO: https://datadoghq.atlassian.net/browse/RUMM-1894
+
+        // Given
+        stubGradleBuildFromResourceFile(
+            "build_with_datadog_dep.gradle.kts",
+            appBuildGradleFile
+        )
+
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments("--configuration-cache", ":samples:app:assembleRelease")
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":samples:app:assembleRelease")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @Ignore(
+        "This test is ignored for now as we are using the Configuration object at the " +
+            "task action level in our DdCheckSdkDepsTask and this is breaking " +
+            "the --configuration-cache. There is no workaround this yet and this is " +
+            "also present in some internal build.gradle tasks (see the test comment above)"
+    )
+    fun `M success W assembleDebug { configuration cache, checkProjectDependecies enabled }`() {
+        // TODO: https://datadoghq.atlassian.net/browse/RUMM-1893
+
+        // Given
+        stubGradleBuildFromResourceFile(
+            "build_with_datadog_dep.gradle.kts",
+            appBuildGradleFile
+        )
+
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments("--configuration-cache", ":samples:app:assembleDebug")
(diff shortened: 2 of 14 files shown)


.gitlab-ci.yml
```diff
@@ -99,6 +99,7 @@ test:plugin:
     - git fetch --depth=1 origin main
     - rm -rf ~/.gradle/daemon/
     - CODECOV_TOKEN=$(aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android-gradle-plugin.codecov-token --with-decryption --query "Parameter.Value" --out text)
+    - GRADLE_OPTS="-XX:MaxPermSize=512m -Xmx2560m" ./gradlew :dd-sdk-android-gradle-plugin:publishToMavenLocal -Pdd-skip-signing
```
0xnm, line 102: Is it necessary?

buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
```diff
@@ -114,7 +114,7 @@ fun Project.publishingConfig(projectDescription: String) {
         signingExtension.apply {
             val privateKey = System.getenv("GPG_PRIVATE_KEY")
             val password = System.getenv("GPG_PASSWORD")
-            isRequired = true
+            isRequired = !hasProperty("dd-skip-signing")
```
0xnm, line 117: should be reverted

gradle/libs.versions.toml
```diff
@@ -42,7 +42,7 @@ kotlinGrammarParser = "c35b50fa44"
 
 # Datadog
 datadogSdk = "1.11.0"
-datadogPluginGradle = "1.2.0"
+datadogPluginGradle = "1.2.1-dev"
```
0xnm, line 45: so to make things work we always need to have plugin in maven local?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,287 @@
...
+    @Test
+    fun `M success W assemble`() {
+        // Given
+        stubGradleBuildFromResourceFile("build_with_datadog_dep.gradle.kts")
+
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments("assemble")
+            .withPluginClasspath()
```
0xnm, line 48: It seems there is no need to publish to the maven local if we use this method? If I get explanation here https://docs.gradle.org/current/javadoc/org/gradle/testkit/runner/GradleRunner.html#withPluginClasspath-- right

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,287 @@
...
+        assertThat(uploadTask).isNull()
+    }
+
+    // endregion
+
+    // region Internal
+
+    private fun stubFile(destination: File, content: String) {
+        with(destination.outputStream()) {
+            this.write(content.toByteArray())
```
0xnm, line 235:
```suggestion
            write(content.toByteArray())
```

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,287 @@
...
+        const val SETTINGS_FILE_CONTENT = "rootProject.name = 'sample'"
+        val GRADLE_PROPERTIES_FILE_CONTENT = """
+           org.gradle.jvmargs=-Xmx2560m -XX:MaxPermSize=512m
+           android.useAndroidX=true
+           DD_API_KEY=fakekey
+        """.trimIndent()
+
+        @Suppress("unused")
+        @JvmStatic
+        fun uploadTasks(): Stream<Arguments> {
```
0xnm, line 277: in case of single method argument method source can just return a list of strings per https://junit.org/junit5/docs/5.3.0-RC1/api/org/junit/jupiter/params/provider/MethodSource.html, no need to use Stream + Argument

mariusc83: Ah cool, didn't know that ;)

buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
mariusc83: I need this here for `publishToMavenLocal` to work. Is this what you were asking ?

.gitlab-ci.yml
mariusc83: Yes, unfortunately was not able to make it work to get the plugin from the classpath. For some reason it doesn't work with AGP. I am going to give it one more shot today.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,287 @@
...
+            .withArguments("--build-cache", "assemble")
+            .withPluginClasspath()
+            .build()
+
+        // Then
+        assertThat(result.task(":assemble")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @Test
+    fun `M fail W assemble { configuration cache, checkProjectDependecies enabled }`() {
```
xgouchet, line 72: Why should the plugin fail in this case? Also can we be a bit more precise in what kind of exception we're expecting ?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,287 @@
...
+            GradleRunner.create()
+                .withProjectDir(testProjectDir)
+                .withArguments("--configuration-cache", "assemble")
+                .withPluginClasspath()
+                .build()
+        }.isInstanceOf(Exception::class.java)
+    }
+
+    // The assembleRelease fails for configuration cache and not because of our code but because of
+    // these tasks: collect[Flavour1][Flavour2]ReleaseDependencies
```
xgouchet, line 87: Is this comment related to the test method above? If so can you move it to the test method body instead?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,287 @@
...
+        // Given
+        stubGradleBuildFromResourceFile("build_without_datadog_dep.gradle.kts")
+        // When
+        assertThatThrownBy {
+            GradleRunner.create()
+                .withProjectDir(testProjectDir)
+                .withArguments("assemble")
+                .withPluginClasspath()
+                .build()
+        }.isInstanceOf(Exception::class.java)
```
xgouchet, line 131: Again can we make a more precise check on which exception we're expecting here?

xgouchet (changes requested):

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: Not so sure as it is a generic exception but will try to have a look

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: Good catch, this one should not fail actually is just that I need to test the `assembleDebug` instead as `assemble` contains `assembleRelease` also and that ones fails for other reasons than our code (see the comment below).

mariusc83: The reason why this test should fail is that our `checkDepsTask` is not aligned with the `--configuration-cache` requirements because we are using the `project` inside. @0xnm tried to fix this a while ago but it seems too tricky for now. We have an entry in our `Readme` file that states this but I will also add an explanation into this test.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,323 @@
...
+        // could not easily solve that.
+
+        // Given
+        stubGradleBuildFromResourceFile("build_with_datadog_dep.gradle.kts")
+
+        // When
+        assertThatThrownBy {
+            GradleRunner.create()
+                .withProjectDir(testProjectDir)
+                .withArguments("--configuration-cache", "assemble")
```
0xnm, line 87: shouldn't we test `assembleDebug` here per test name?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,323 @@
...
+        // Given
+        stubGradleBuildFromResourceFile("build_without_datadog_dep.gradle.kts")
+        // When
+        assertThatThrownBy {
+            GradleRunner.create()
+                .withProjectDir(testProjectDir)
+                .withArguments("assemble")
+                .withPluginClasspath(getTestConfigurationClasspath())
+                .build()
+        }.isInstanceOf(UnexpectedBuildFailure::class.java)
```
0xnm, line 137: should we check that cause is `MissingSdkException` or it is not present in the exception thrown?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,323 @@
...
+            .build()
+
+        // Then
+        assertThat(result.task(":$taskName")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @ParameterizedTest
+    @MethodSource("uploadTasks")
+    fun `M success W upload { checkProjectDependecies disabled }`(taskName: String) {
```
0xnm, line 182: I think this `checkProjectDependencies` has no impact on the upload task, so we can skip tests with that maybe?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,323 @@
...
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":$taskName")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    @Test
+    fun `M not contain any uploadTasks W minifyNotEnabeld`() {
```
0xnm, line 217:
```suggestion
    fun `M not contain any uploadTasks W minifyNotEnabled`() {
```

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,323 @@
...
+    @ParameterizedTest
+    @MethodSource("uploadTasks")
+    fun `M success W upload`(taskName: String) {
+        // Given
+        stubGradleBuildFromResourceFile("build_with_datadog_dep.gradle.kts")
+
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments(taskName)
```
0xnm, line 153: without calling `assembleXXX` there won't be mapping file and task will be effectively no-op according to this line https://github.com/DataDog/dd-sdk-android-gradle-plugin/blob/ba2d61297317a004efab56665257dcdbaff9e0be/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt#L106.

Mapping Task is independent and doesn't call `assemble` by itself.

Also maybe we need to add real files/modules to the `src/main`? Not sure, but it it might be that without our files there won't be any mapping file.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,323 @@
...
+        manifestFile = File(mainSrcDir, "AndroidManifest.xml")
+        stubFile(settingsFile, SETTINGS_FILE_CONTENT)
+        stubFile(manifestFile, ANDROID_MANIFEST_FILE_CONTENT)
+        stubFile(gradlePropertiesFile, GRADLE_PROPERTIES_FILE_CONTENT)
+    }
+
+    // region Assemble
+
+    @Test
+    fun `M success W assemble`() {
```
0xnm, line 41: can we try to add the test (well, project structure with app module with flavors + lib module without flavors) which will check that regression of this bug https://github.com/DataDog/dd-sdk-android-gradle-plugin/issues/62 doesn't happen?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,323 @@
...
+
+    companion object {
+        val ANDROID_MANIFEST_FILE_CONTENT = """
+            <?xml version="1.0" encoding="utf-8"?>
+            <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+                      package="com.example.variants">
+
+                <application
+                    android:allowBackup="true"
+                    android:supportsRtl="true">
```
0xnm, line 272: let's add also `exported=true` here, so that compilation succeeds for API 31 (I will handle the rest in the https://github.com/DataDog/dd-sdk-android-gradle-plugin/pull/53)

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,323 @@
...
+
+            </manifest>
+        """.trimIndent()
+
+        const val SETTINGS_FILE_CONTENT = """
+            pluginManagement {
+                resolutionStrategy {
+                    eachPlugin {
+                        if (requested.id.id == "com.android.application") {
+                            useModule("com.android.tools.build:gradle:7.0.3")
```
0xnm, line 290: ideally versions here should be taken from versions catalog and be the same as we use in the plugin dependencies, but tbh I don't know how to reference them.

Maybe some property which we can inject from `buildSrc` into test environment? Same for the DD SDK version in the test build script files.

0xnm (changes requested):
Awesome work! I left few comments about how can we cover more and be more flexible in the test setup.

xgouchet (changes requested):

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: Actually I just wanted to explain why we cannot provide relevant tests for `assembleRelease` . I should probably state this outside of the function when the region starts.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: Good catch.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: Could not see it in the thrown exception but will have another look.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: There is more than that actually, it needs to be a git versioned directory otherwise it will not work (I think you already did something like that for `GitRepositoryDetectorTest`). I will have a look into this and try to provide a dummy Application class in the `src/main`

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: Ok, I will try to see how we can achieve this.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: Yeah, this is going to be tricky I guess. I am thinking that maybe we can do this in another task as there are already several other things to tackle into this one.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: Unfortuantely I cannot get the root cause of this exception as `GradleRunner` is intercepting that but it is not passing it along :(

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
mariusc83: As I explained also for Xavier's comment we are not able to get the root exception as GradleRunner is actually intercepting it under the hood but not passing it along with the result or in the thrown UnexpectedBuildFailure.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,511 @@
...
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":samples:app:assembleDebug")?.outcome)
+            .isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    // The assembleRelease fails for configuration cache and not because of our
+    // code but because of these tasks: collect[Flavour1][Flavour2]ReleaseDependencies. For this
```
xgouchet, line 207: What is making those tasks fail, then?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,511 @@
...
+    // reason the below 2 scenarios are only tested for the `debug` build type.
+
+    @Test
+    fun `M fail W assembleDebug { configuration cache, checkProjectDependecies enabled }`() {
+
+        // This test should fail because our `DdCheckSdkDepsTask` is not following the
+        // --configuration-cache best practices (we are using the project inside) and so far we
+        // could not easily solve that.
```
xgouchet, line 215: Wait, you mean that for now the `assembleDebug` fails but because there's a bug in our project, and not because it's expected to fail? Then we shouldn't have this test at all. We need to write prescriptive tests, not descriptive ones.
Put in other word: 
- if we want the `assembleDebug` task to fail in this specific scenario (using the `build_with_datadog_dep.gradle.kts` file and running the task with `--configuration-cache`), then yes we can keep this test;
- if we want the task to succeed, but it fails for technical reasons (untestable, non compliance on our part with the configuration cache or anything else), then either we make it succeed in the test or we do write the test for success, mark it as `@Ignored` and create a task in JIRA to make it pass.

xgouchet (changes requested):

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,504 @@
...
+        )
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments(":samples:app:assembleRelease")
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        print(result.output)
```
0xnm, line 102: should be removed

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,504 @@
...
+
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments(":samples:app:assembleRelease")
+            .withPluginClasspath(getTestConfigurationClasspath())
+            .build()
+
+        // Then
+        assertThat(result.task(":samples:app:assembleRelease")?.outcome)
```
0xnm, line 266: should we try also to assert that output contains a warning?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -0,0 +1,504 @@
...
+            "build_with_datadog_dep.gradle.kts",
+            appBuildGradleFile
+        )
+        val taskName = forge.anElementFrom(uploadTasks)
+        val variantName = taskName.extractVariantName()
+
+        // When
+        val result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .withArguments(":samples:app:assembleRelease", taskName, "--info")
```
0xnm, line 338: nice workaround!

0xnm (approved):
I left few minor comments, but overall lgtm!

xgouchet (approved):

Issue #71, opened 2022-02-07
Add param to customise mapping file path for upload task


pietrini:
**Is your feature request related to a problem? Please describe.**

We use DexGuard to obfuscate our code. Unfortunately, the plugin uses a dedicated folder and not the usual one, mapping file will be in `/build/outputs/dexguard/mapping/bundle/release/mapping.txt`.

**Describe the solution you'd like**

    datadog {
      ...
      mappingFile = ".../mapping.txt"
      ...
    }

**Describe alternatives you've considered**

I'm launching a task to copy my mapping file in the expected folder before launching the upload mapping task.

PR #72, opened 2022-02-18
RUMM-1972 Provide mappingFilePath through plugin configuration

mariusc83:
### What does this PR do?

Adds a new property for the plugin extension configuration -> `mappingFilePath`. Using this new property users will be able to provide their own custom `mappingFilePath` for each application variant in case they are using any other third party code obfuscator than the default one.

Close issue #71 

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
@@ -91,10 +91,8 @@ class DdAndroidGradlePlugin @Inject constructor(
         configureVariantTask(uploadTask, apiKey, flavorName, extensionConfiguration, variant)
 
         val outputsDir = File(target.buildDir, "outputs")
-        val mappingDir = File(outputsDir, "mapping")
-        val flavorDir = File(mappingDir, variant.name)
-        uploadTask.mappingFilePath = File(flavorDir, "mapping.txt").path
-
+        uploadTask.mappingFilePath =
+            resolveMappingFilePath(extensionConfiguration, outputsDir, variant)
         val reportsDir = File(outputsDir, "reports")
         val datadogDir = File(reportsDir, "datadog")
         uploadTask.repositoryFile = File(datadogDir, "repository.json")
@@ -155,6 +153,21 @@ class DdAndroidGradlePlugin @Inject constructor(
             }
             compileTask.finalizedBy(checkDepsTaskProvider)
             return checkDepsTaskProvider
+        }
+    }
+
+    private fun resolveMappingFilePath(
+        extensionConfiguration: DdExtensionConfiguration,
+        outputsDir: File,
+        variant: ApplicationVariant
+    ): String {
+        val customPath = extensionConfiguration.mappingFilePath
+        return if (customPath != null) {
+            customPath
+        } else {
+            val mappingDir = File(outputsDir, "mapping")
+            val flavorDir = File(mappingDir, variant.name)
+            File(flavorDir, "mapping.txt").path
         }
     }
 
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtensionConfiguration.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtensionConfiguration.kt
@@ -40,11 +40,18 @@ open class DdExtensionConfiguration(
      */
     var checkProjectDependencies: SdkCheckLevel? = null
 
+    /**
+     * The absolute path to the mapping file to be used for deobfuscation. If not provided the
+     * default one will be used: [buildDir]/output/mapping/[variant]/mapping.txt
+     */
+    var mappingFilePath: String? = null
+
     internal fun updateWith(config: DdExtensionConfiguration) {
         config.versionName?.let { versionName = it }
         config.serviceName?.let { serviceName = it }
         config.site?.let { site = it }
         config.remoteRepositoryUrl?.let { remoteRepositoryUrl = it }
         config.checkProjectDependencies?.let { checkProjectDependencies = it }
+        config.mappingFilePath?.let { mappingFilePath = it }
     }
 }
--- a/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
+++ b/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
@@ -104,10 +104,9 @@ internal class DdAndroidGradlePluginTest {
         assertThat(task.variantName).isEqualTo(flavorName)
         assertThat(task.versionName).isEqualTo(versionName)
         assertThat(task.serviceName).isEqualTo(packageName)
-        assertThat(task.remoteRepositoryUrl).isEmpty()
         assertThat(task.site).isEqualTo(fakeExtension.site)
-        assertThat(task.mappingFilePath)
-            .isEqualTo("${fakeProject.buildDir}/outputs/mapping/$variantName/mapping.txt")
+        assertThat(task.remoteRepositoryUrl).isEqualTo(fakeExtension.remoteRepositoryUrl)
+        assertThat(task.mappingFilePath).isEqualTo(fakeExtension.mappingFilePath)
     }
 
     @Test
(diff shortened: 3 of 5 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -158,6 +156,21 @@ class DdAndroidGradlePlugin @Inject constructor(
         }
     }
 
+    private fun resolveMappingFilePath(
+        extensionConfiguration: DdExtensionConfiguration,
+        outputsDir: File,
+        variant: ApplicationVariant
+    ): String {
+        val customPath = extensionConfiguration.mappingFilePath
```
xgouchet, line 164: You can't read the extension directly here. When this method is called, you're still configuring the project instance (you're in the `apply plugin` part of the project's `build.gradle` script). 
This means that the value of the `mappingFilePath` will always be the default. 

What you need to do is use the `project.afterEvaluate { }` block to wait until the full `build.gradle` script has been read, and the `mappingFilePath` has been set by the customer.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtensionConfiguration.kt
```diff
@@ -40,11 +40,18 @@ open class DdExtensionConfiguration(
      */
     var checkProjectDependencies: SdkCheckLevel? = null
 
+    /**
+     * The absolute path to the mapping file to be used for deobfuscation. If not provided the
+     * default one will be used: [buildDirectoryPath]/output/mapping/[variant]/mapping.txt
```
xgouchet, line 45:
```suggestion
     * default one will be used: [buildDir]/output/mapping/[variant]/mapping.txt
```

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
```diff
@@ -145,10 +144,9 @@ internal class DdAndroidGradlePluginTest {
         assertThat(task.variantName).isEqualTo(flavorName)
         assertThat(task.versionName).isEqualTo(fakeExtension.versionName)
         assertThat(task.serviceName).isEqualTo(fakeExtension.serviceName)
-        assertThat(task.remoteRepositoryUrl).isEmpty()
         assertThat(task.site).isEqualTo(fakeExtension.site)
-        assertThat(task.mappingFilePath)
-            .isEqualTo("${fakeProject.buildDir}/outputs/mapping/$variantName/mapping.txt")
+        assertThat(task.remoteRepositoryUrl).isEqualTo(task.remoteRepositoryUrl)
```
xgouchet, line 148: Are you sure you want to test that `task.remoteRepositoryUrl` is equal to itself?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
```diff
@@ -104,10 +104,9 @@ internal class DdAndroidGradlePluginTest {
         assertThat(task.variantName).isEqualTo(flavorName)
         assertThat(task.versionName).isEqualTo(versionName)
         assertThat(task.serviceName).isEqualTo(packageName)
-        assertThat(task.remoteRepositoryUrl).isEmpty()
         assertThat(task.site).isEqualTo(fakeExtension.site)
-        assertThat(task.mappingFilePath)
-            .isEqualTo("${fakeProject.buildDir}/outputs/mapping/$variantName/mapping.txt")
+        assertThat(task.remoteRepositoryUrl).isEqualTo(task.remoteRepositoryUrl)
```
xgouchet, line 108: Are you sure you want to test that `task.remoteRepositoryUrl` is equal to itself?

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdExtensionForgeryFactory.kt
```diff
@@ -18,6 +18,12 @@ internal class DdExtensionForgeryFactory : ForgeryFactory<DdExtension> {
             site = forge.aValueFrom(DatadogSite::class.java).name
             variants = mock()
             checkProjectDependencies = forge.aValueFrom(SdkCheckLevel::class.java)
+            remoteRepositoryUrl = forge.aStringMatching(
+                "https://[a-z]{4,10}\\.[com|org]/[a-z]{4,10}/[a-z]{4,10}\\.git"
+            )
+            mappingFilePath = forge.aStringMatching(
+                "([a-z]+)/([a-z]+)/([a-z]+)/mapping.txt"
+            )
```
xgouchet, line 26:
```suggestion
            mappingFilePath = forge.aNullable {
                aStringMatching("([a-z]+)/([a-z]+)/([a-z]+)/mapping.txt")
            }
```

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdExtensionConfigurationForgeryFactory.kt
```diff
@@ -19,6 +19,9 @@ internal class DdExtensionConfigurationForgeryFactory : ForgeryFactory<DdExtensi
                 "https://[a-z]{4,10}\\.[com|org]/[a-z]{4,10}/[a-z]{4,10}\\.git"
             )
             checkProjectDependencies = forge.aValueFrom(SdkCheckLevel::class.java)
+            mappingFilePath = forge.aStringMatching(
+                "([a-z]+)/([a-z]+)/([a-z]+)/mapping.txt"
+            )
```
xgouchet, line 24:
```suggestion
            mappingFilePath = forge.aNullable {
                aStringMatching("([a-z]+)/([a-z]+)/([a-z]+)/mapping.txt")
            }
```

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
```diff
@@ -145,10 +144,9 @@ internal class DdAndroidGradlePluginTest {
         assertThat(task.variantName).isEqualTo(flavorName)
         assertThat(task.versionName).isEqualTo(fakeExtension.versionName)
         assertThat(task.serviceName).isEqualTo(fakeExtension.serviceName)
-        assertThat(task.remoteRepositoryUrl).isEmpty()
         assertThat(task.site).isEqualTo(fakeExtension.site)
-        assertThat(task.mappingFilePath)
-            .isEqualTo("${fakeProject.buildDir}/outputs/mapping/$variantName/mapping.txt")
+        assertThat(task.remoteRepositoryUrl).isEqualTo(task.remoteRepositoryUrl)
+        assertThat(task.mappingFilePath).isEqualTo(fakeExtension.mappingFilePath)
```
xgouchet, line 149: You need to build the `expectedPath` to handle both cases when the mapping file uses the default or when it's set by the customer.

xgouchet (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
mariusc83: this function : `configureVariantForUploadTask` is being called in the `target.afterEvaluate { }` block above in the code so the value will already be there. I also tested it on my end and it works.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
mariusc83: No, that was a mistake ;)

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
mariusc83: Same as above

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdExtensionForgeryFactory.kt
mariusc83: Yes I know but I wanted to follow the pattern as we are using non - nullable generators for all the other properties in the extension (even though they are nullable). If we want to to use this here then it will make sense to change it for all the others and adapt the tests. I did not think this should be the topic for this PR.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTest.kt
mariusc83: I have 2 tests for that: 
`𝕄 configure the upload task with the extension info 𝕎 configureVariant()` and `𝕄 use sensible defaults 𝕎 configureVariant() { empty config }`. The only problem I see here is that they are testing all the properties, if we chose to have dedicated tests for the `mappingFilePath` then I say we should have dedicated tests for each property in the `DdExtension` class.

xgouchet (approved):

PR #86, opened 2022-05-06
RUMM-2174: Add shrinking options for Proguard/R8 mapping file

0xnm:
### What does this PR do?

This PR allows to shrink a `mapping.txt` file in the following way:

1) Use short aliases for packages/full class names for source items of the `mapping.txt` file. For example, `androidx.appcompat` can be replaced by `axapp` if needed.

This allows to reduce a size of the mapping file a bit (but napkin math shows that this reduction will be no more than just few percent), allowing to proceed with uploading mapping file which slightly exceeds 50 Mb limit. It is a quite specialized workaround, so we don't document the field controlling this feature.

The effort is done to check that names coming from the application package are not renamed to keep the integration with VCS, only 3rd party names are allowed to be changed.

The field controlling this feature doesn't inherit the values from the parent declaration in order to be explicit. It means that if configuration if declared for the particular variant, aliasing rules should be declared there as well **completely**, even if it is already declared in the parent config.

2) Trim line indents, they are not used during `mapping` parsing anyway.


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
```diff
@@ -179,5 +193,61 @@ open class DdMappingFileUploadTask
...
+            }
+
+        return shrinkedFile
+    }
+
+    private fun applyShortAliases(
+        line: String,
+        replacements: List<Pair<Regex, String>>
+    ): String {
+        val isLineWithRename = !line.startsWith(MAPPING_FILE_COMMENT_CHAR) &&
```
daniel-m-dd, line 231: so we can dont rename `synthetized` comment even if we are renaming the fields on which the comment apply ?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
```diff
@@ -179,5 +193,61 @@ open class DdMappingFileUploadTask
...
+            line.contains(MAPPING_FILE_CHANGE_DELIMITER)
+
+        return if (isLineWithRename) {
+            val (source, obfuscated) = line.split(MAPPING_FILE_CHANGE_DELIMITER)
```
daniel-m-dd, line 235: shouldnt we check on the number of `MAPPING_FILE_CHANGE_DELIMITER` in the string (if > 1?)

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
```diff
@@ -179,5 +193,61 @@ open class DdMappingFileUploadTask
...
+            val aliasedSource =
+                replacements.fold(source) { state, entry ->
+                    val (from, to) = entry
+                    state.replace(from, to)
```
daniel-m-dd, line 239: do we have benefits/use case allowing more than 1 replacement per line ?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
daniel-m-dd: or maybe use regex with group matching ?

daniel-m-dd (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
0xnm: `Synthesized` comment doesn't carry the name, it is just a directive placed below the frame saying that it should be removed from retraced stack according to https://r8.googlesource.com/r8/+/refs/heads/main/doc/retrace.md#synthesized-introduced-at-version-1_0.

It seems only `RewriteFrame` comment can carry the name (in the `conditions` field), but I didn't find such comments in the mapping file of Datadog app, so I guess they are pretty rare, and even if we renamed something which stayed non-renamed in the `conditions`, worst case scenario will be just that `RewriteFrame` directive won't be applied.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
0xnm: Seems it is not possible to have multiple `->` on the line with the current format, but indeed, I will add a check just in case.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
0xnm: We need to do that because it can be several types as function arguments mentioned on the single line, so several replacements can apply, ex.:

> 39:39:void ModalBottomSheetLayout-BzaUkTc(kotlin.jvm.functions.Function3,androidx.compose.ui.Modifier,androidx.compose.material.ModalBottomSheetState,androidx.compose.ui.graphics.Shape,float,long,long,long,kotlin.jvm.functions.Function2,androidx.compose.runtime.Composer,int,int):392:392 -> a

xgouchet (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtensionConfiguration.kt
```diff
@@ -47,12 +47,30 @@ open class DdExtensionConfiguration(
...
+     * ```
+     *  "androidx.appcompat" to "axapp",
+     *  "androidx.work" to "axw",
+     *  "java.lang" to "jl",
+     *  "kotlin.collections" to "kc"
+     *  ...
+     * ```
+     *
+     * This property is not inherited, meaning in the particular variant configuration it should
+     * be declared explicitly, even if it exists in the root configuration.
```
xgouchet, line 64: Is there a limitation preventing us to inherit this for variant? IMHO this would make it tedious to copy the same conf in all variants.

xgouchet (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdExtensionConfiguration.kt
0xnm: The motivation behind that is the following:

* Not necessarily a particular variant needs that, it may have `mapping.txt` which simply fits the limit, so we shouldn't blindly apply what was declared in the root configuration (in the end introducing short aliases impacts the readability of the retraced stacks).
* If we say that variant should inherit this property, then what is the approach for inheritance: do we replace the mapping? Or do we merge what is declared in the variant with what is declared in the root configuration?

So considering that I prefer to not have an inheritance. In the end if user needs to share some mapping between variants, they can simply introduce a method which adds it with the format and merging strategy they want.

mariusc83 (commented):
Maybe we should add some tests also into the `DdAndroidGradlePluginFunctionalTest`. Mostly for those 2 options added in the plugin configuration to make sure that everything works well.

mariusc83 (approved):

xgouchet (approved):

PR #99, opened 2022-07-19
RUMM-2342 Add support to v2 request

xgouchet:
### What does this PR do?

Adapt the payload format to API v2

### Motivation

Stay compatible with the backend processor evolutions

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
@@ -8,6 +8,9 @@ package com.datadog.gradle.plugin.internal
 
 import com.datadog.gradle.plugin.DdAndroidGradlePlugin.Companion.LOGGER
 import com.datadog.gradle.plugin.RepositoryInfo
+import java.io.File
+import java.net.HttpURLConnection
+import java.util.concurrent.TimeUnit
 import okhttp3.MediaType.Companion.toMediaTypeOrNull
 import okhttp3.MultipartBody
 import okhttp3.OkHttpClient
@@ -16,9 +19,6 @@ import okhttp3.RequestBody.Companion.asRequestBody
 import okhttp3.RequestBody.Companion.toRequestBody
 import okhttp3.Response
 import org.json.JSONObject
-import java.io.File
-import java.net.HttpURLConnection
-import java.util.concurrent.TimeUnit
 
 internal class OkHttpUploader : Uploader {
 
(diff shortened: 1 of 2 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
```diff
@@ -81,12 +88,16 @@ internal class OkHttpUploader : Uploader {
...
+        eventJson.put("type", TYPE_JVM_MAPPING_FILE)
+
         val builder = MultipartBody.Builder()
         builder.setType(MultipartBody.FORM)
-            .addFormDataPart("version", identifier.version)
-            .addFormDataPart("service", identifier.serviceName)
-            .addFormDataPart("variant", identifier.variant)
-            .addFormDataPart("type", TYPE_JVM_MAPPING_FILE)
+            .addFormDataPart("event", "event", eventJson.toString(0).toRequestBody(MEDIA_TYPE_JSON))
             .addFormDataPart("jvm_mapping_file", mappingFile.name, mappingFileBody)
```
rpelliard, line 101:
```suggestion
            .addFormDataPart("jvm_mapping_file", "jvm_mapping", mappingFileBody)
```
Could we set a consistent `jvm_mapping` file name here ? This is required by the v2 intake to correctly process the blob

rpelliard: And same for the repository file, I think it needs `repository` as a filename (the v2 intake relies on file names instead of data part names)

0xnm (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
```diff
@@ -137,13 +162,17 @@ internal class OkHttpUploader : Uploader {
         internal const val HEADER_EVP_ORIGIN_VERSION = "DD-EVP-ORIGIN-VERSION"
         internal const val HEADER_REQUEST_ID = "DD-REQUEST-ID"
 
+        internal const val KEY_EVENT = "event"
+        internal const val KEY_JVM_MAPPING_FILE = "jvm_mapping_file"
```
rpelliard, line 166:
```suggestion
        internal const val KEY_JVM_MAPPING_FILE = "jvm_mapping"
```
The file name needs to be exactly `jvm_mapping` (the part name doesn't really matter), other than that LGTM !

rpelliard (approved):
👍

PR #104, opened 2022-07-28
RUMM-2345: Read config properties from datadog-ci.json

0xnm:
### What does this PR do?

This change supports parsing some properties from `datadog-ci.json`, namely `datadogSite` and `apiKey` properties.

This file may be located in the root of the project (but since the definition of the root may vary between the different setups, we try to find `datadog-ci.json` file by doing upward traversal).

The order of API key lookup is like the following:

* Gradle property
* `datadog-ci.json` file
* Environment property

The order of site lookup is like the following:

* Gradle extension configuration
* `datadog-ci.json` file

`datadog-ci.json` file is included as an input of upload task so that it can be watched by the Gradle for the changes, otherwise it won't be checked and any changes there won't be reflected when running upload task, which may be confusing for the user.


jtappa (approved):

mariusc83 (approved):
LGTM

xgouchet (approved):
Nice work 👍

PR #119, opened 2022-10-13
Use shared CI template

daniel-m-dd:
### What does this PR do?

Use a shared template for some CI jobs.

### Motivation

### Additional Notes


build.gradle.kts
```diff
@@ -34,3 +33,5 @@ allprojects {
 task<Delete>("clean") {
     delete(rootProject.buildDir)
 }
+
+tasks.register("checkGeneratedFiles") {}
```
0xnm, line 37: seems it is some leftover?

samples/basic/src/main/java/com/example/variants/MainActivity.kt
```diff
@@ -3,7 +3,6 @@ package com.example.variants
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 // unused import is on purpose
-import com.example.lib.Placeholder
```
0xnm, line 6: we shouldn't remove this line

samples/variants/src/main/java/com/example/variants/MainActivity.kt
```diff
@@ -1,9 +1,8 @@
 package com.example.variants
 
+// unused import is on purpose
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
-// unused import is on purpose
-import com.example.lib.Placeholder
```
0xnm, line 6: same - shouldn't remove this line

0xnm (changes requested):

build.gradle.kts
daniel-m-dd: it is actually needed by the shared pipeline (but no use here so it is a task doing nothing)

0xnm: can we then add a comment? so that empty task is not confusing

0xnm (approved):

xgouchet (approved):

PR #131, opened 2023-01-10
RUMM-2770: Improve mapping upload flow handling

0xnm:
### What does this PR do?

This change improves mapping upload flow handling by doing the following:

* Additional check for the API key validity is done (using [API key validation API](https://docs.datadoghq.com/api/latest/authentication/#validate-api-key)) if server returns code 400, this is basically aligns logic with [datadog-ci](https://github.com/DataDog/datadog-ci/blob/ffc52ea6c0b0ac44004544bf84689a22ce2d4119/src/helpers/apikey.ts#L48).
* Removed a statement saying that server may already have mapping uploaded - server will never return error if there is already mapping uploaded.
* Added an additional check for quotes\apostrophes in API key, this may be useful for the error investigation on the client side (this check is quite limited to only quotes and apostrophes, because there is no public contract for the API key format).
* Added a site name to the upload info message (and to some other messages as well).

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
@@ -136,6 +136,12 @@ open class DdMappingFileUploadTask
         applySiteFromEnvironment()
         validateConfiguration()
 
+        if (apiKey.contains("\"") || apiKey.contains("'")) {
+            throw IllegalStateException(
+                "DD-API-KEY provided shouldn't contain quotes or apostrophes."
+            )
+        }
+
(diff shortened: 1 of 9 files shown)


xgouchet (approved):
Nice work!

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
```diff
@@ -136,6 +136,12 @@ open class DdMappingFileUploadTask
         applySiteFromEnvironment()
         validateConfiguration()
 
+        if (apiKey.contains("\"") || apiKey.contains("'")) {
```
mariusc83, line 139: maybe you should extract this in a method for better readability

mariusc83 (approved):
LGTM, left one comment there

PR #141, opened 2023-02-02
Improve the documentation for 50Mb limit

xgouchet:
### What does this PR do?

Improve the documentation to give workarounds when the mapping file is above 50 Mb

### DIFF
--- a/docs/upload_mapping_file.md
+++ b/docs/upload_mapping_file.md
@@ -146,7 +146,7 @@ than 50Mb, you can try several options to reduce the file size and be able to up
 
 ```
 datadog {
-    mappingFileTrimIndents = false
+    mappingFileTrimIndents = true
     mappingFilePackageAliases = mapOf(
         "kotlinx.coroutines" to "kx.cor",
         "com.google.android.material" to "material",
--- a/docs/upload_mapping_file.md
+++ b/docs/upload_mapping_file.md
@@ -136,6 +136,26 @@ For example:
 tasks["minify${variant}WithR8"].finalizedBy { tasks["uploadMapping${variant}"] }
 ```
 
+## Limitations
+
+For now, we can only upload mapping files with a size smaller than 50 Mb. If your project has a mapping file larger
+than 50Mb, you can try several options to reduce the file size and be able to upload it.
+
+- set the `mappingFileTrimIndents` option to true, this will reduce your file size by 5% on average
+- set a map of `mappingFilePackagesAliases`: this will replace package names by shorter alias. The main caveat is that the stacktrace in Datadog will use the same alias instead of the original package name, so it's better to use it for third party dependencies.
+
+```
+datadog {
+    mappingFileTrimIndents = false
+    mappingFilePackageAliases = mapOf(
+        "kotlinx.coroutines" to "kx.cor",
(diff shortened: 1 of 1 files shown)


docs/upload_mapping_file.md
```diff
@@ -136,6 +136,26 @@ For example:
...
+
+For now, we can only upload mapping files with a size smaller than 50 Mb. If your project has a mapping file larger
+than 50Mb, you can try several options to reduce the file size and be able to upload it.
+
+- set the `mappingFileTrimIndents` option to true, this will reduce your file size by 5% on average
+- set a map of `mappingFilePackagesAliases`: this will replace package names by shorter alias. The main caveat is that the stacktrace in Datadog will use the same alias instead of the original package name, so it's better to use it for third party dependencies.
+
+```
+datadog {
+    mappingFileTrimIndents = false
```
0xnm, line 149: Should it be `true` in this particular example?

```suggestion
    mappingFileTrimIndents = true
```

0xnm (approved):

docs/upload_mapping_file.md
```diff
@@ -136,6 +136,26 @@ For example:
 tasks["minify${variant}WithR8"].finalizedBy { tasks["uploadMapping${variant}"] }
 ```
 
+## Limitations
+
+For now, we can only upload mapping files with a size smaller than 50 Mb. If your project has a mapping file larger
+than 50Mb, you can try several options to reduce the file size and be able to upload it.
```
cswatt, line 142: Mapping files are limited to 50 Mb. If your project has a mapping file larger than 50 Mb, use one of the following options to reduce the file size:

docs/upload_mapping_file.md
```diff
@@ -136,6 +136,26 @@ For example:
...
+
+- set the `mappingFileTrimIndents` option to true, this will reduce your file size by 5% on average
```
cswatt, line 144:
```suggestion
- Set the `mappingFileTrimIndents` option to `true`. This reduces your file size by 5%, on average.
```

docs/upload_mapping_file.md
```diff
@@ -136,6 +136,26 @@ For example:
...
+- set a map of `mappingFilePackagesAliases`: this will replace package names by shorter alias. The main caveat is that the stacktrace in Datadog will use the same alias instead of the original package name, so it's better to use it for third party dependencies.
```
cswatt, line 145:
```suggestion
- Set a map of `mappingFilePackagesAliases`: This replaces package names with shorter aliases. **Note**: Datadog's stacktrace uses the same alias instead of the original package name, so it's better to use this option for third party dependencies.
```

cswatt (changes requested):

docs/upload_mapping_file.md
xgouchet, line 142:
```suggestion
Mapping files are limited to 50 Mb. If your project has a mapping file larger than 50 Mb, use one of the following options to reduce the file size:
```

cswatt (approved):

plousada:
We now support 100Mb in US1, should we change this doc?

xgouchet:
@plousada yes that would make sense, I'll update the PR today

PR #152, opened 2023-02-22
Fix CI warnings and issues

xgouchet:


buildSrc/build.gradle.kts
```diff
@@ -61,10 +59,6 @@ gradlePlugin {
     }
 }
 
-tasks.withType<KotlinCompile> {
```
0xnm, line 64: we don't need it anymore?

samples/variants/build.gradle
```diff
@@ -1,6 +1,6 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
-//apply plugin: 'com.datadoghq.dd-sdk-android-gradle-plugin'
```
0xnm, line 3: it seems this change should be reverted, this is a valid id per https://github.com/DataDog/dd-sdk-android-gradle-plugin/blob/0c19f89d19fbbe1de91cd0580b0f2b9fc3cdbf1b/dd-sdk-android-gradle-plugin/build.gradle.kts#L70

samples/variants/src/main/java/com/datadog/example/variants/MainActivity.kt
```diff
@@ -1,9 +1,8 @@
-package com.example.variants
+package com.datadog.example.variants
 
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
-import com.example.lib.Placeholder // ktlint-disable no-unused-imports  unused import is on purpose
```
0xnm, line 5: need to revert this line

buildSrc/build.gradle.kts
xgouchet: No, it actually break the ci as the jvm running gradle uses JDK 17

0xnm (approved):

PR #166, opened 2023-05-09
RUMM-3267: Remove mapping file size check

0xnm:
### What does this PR do?

Requests too big will fail with status 413 and message `Request too large`, so removing mapping file size check from the plugin to rely on the status code instead. Mapping file size limits are mentioned in the docs.

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
@@ -28,6 +28,7 @@ import org.json.JSONTokener
 import java.io.File
 import java.io.IOException
 import java.net.HttpURLConnection
+import java.util.Locale
 import java.util.concurrent.TimeUnit
 
 internal class OkHttpUploader : Uploader {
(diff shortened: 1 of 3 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
```diff
@@ -91,25 +92,12 @@ internal class OkHttpUploader : Uploader {
...
         repositoryInfo: RepositoryInfo?
     ): MultipartBody {
         val mappingFileBody = mappingFile.asRequestBody(MEDIA_TYPE_TXT)
-        // have to use name, because comparison of enum vs spied/mocked enum directly will always fail
-        val sizeLimit = if (site.name == DatadogSite.US1.name) {
-            MAX_MAP_FILE_SIZE_IN_BYTES_US1
-        } else {
-            MAX_MAP_FILE_SIZE_IN_BYTES
-        }
-        if (mappingFileBody.contentLength() > sizeLimit) {
```
emartynov, line 107: Logical and clear change. The only thingy that end user UX is degraded now - I have to wait until the request finishes to find out that it was too big for upload.

PR #220, opened 2024-01-30
feat: Add tasks for upload NDK symbol files to Datadog

fuzzybinary:
### What does this PR do?

This adds tasks for each variant, `ddUploadNdkSymbolFiles{Variant}` to upload symbols from native artifacts to Datadog. It utilizes similar logic from `DdMappingFileUpload` but encapsulates it into `DdFileUploadTask`. This logic is duplicated for now but we should probably refactor `DdMappingFileUpload` to use the `DdFileUploadTask` base class at some point.

NDK Symbol files have the key `ndk_symbol_file`, type `ndk_symbol_file`, and retain their original file name when uploading. The architecture of the symbol file is encoded into the "additional info" map when uploading.

### Additional Notes


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
```diff
@@ -0,0 +1,314 @@
...
+import javax.inject.Inject
+
+/**
+ * A Gradle task to upload symbolication files to Datadog servers (symbol files, proguard files, etc.).
+ */
+abstract class DdFileUploadTask @Inject constructor(
+    providerFactory: ProviderFactory
+) : DefaultTask() {
+    @get:Internal
+    internal var uploader: Uploader = OkHttpUploader()
```
mariusc83, line 33: why not protected ?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -80,10 +80,10 @@ class DdAndroidGradlePlugin @Inject constructor(
         variant: ApplicationVariant,
         apiKey: ApiKey
     ) {
-        if (isObfuscationEnabled(variant, datadogExtension)) {
-            val buildIdGenerationTask =
-                configureBuildIdGenerationTask(target, androidExtension, variant)
+        val buildIdGenerationTask =
```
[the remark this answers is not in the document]
fuzzybinary: Yeup exactly. Before it was only needed if obfuscation was enabled because that's the only way you'd get a mapping file. Now, you can get NDK symbol files even if obfuscation isn't enabled.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
fuzzybinary: Overwritten in unit tests.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
```diff
@@ -0,0 +1,314 @@
...
+        try {
+            val config = JSONObject(datadogCiFile.readText())
+            applyApiKeyFromDatadogCiConfig(config)
+            applySiteFromDatadogCiConfig(config)
+        } catch (e: JSONException) {
+            DdAndroidGradlePlugin.LOGGER.error("Failed to parse Datadog CI config file.", e)
+        }
+    }
+
+    private fun applyApiKeyFromDatadogCiConfig(config: JSONObject) {
```
mariusc83, line 226: I think to be cleaner we can apply the same logic as for the site below to maintain a pattern.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
```diff
@@ -0,0 +1,314 @@
...
+            data.put(it.toJson())
+            DdAndroidGradlePlugin.LOGGER.info(
+                "Detected repository:\n${it.toJson().toString(
+                    INDENT
+                )}"
+            )
+        }
+
+        val jsonObject = JSONObject()
+        jsonObject.put("version", 1)
```
mariusc83, line 288: why do we hardcode the version here ?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * A Gradle task to upload NDK symbol files to Datadog servers.
+ */
+internal abstract class DdSymbolFileUploadTask @Inject constructor(
+    objects: ObjectFactory,
+    providerFactory: ProviderFactory
+) : DdFileUploadTask(providerFactory) {
+
+    @get:InputFiles
+    val searchDirectories: ConfigurableFileCollection = objects.fileCollection()
```
mariusc83, line 28: I see that this property is initialized at registration. I wonder why do you also have to initialize it by default from constructor property. Is this a gradle task requirement ?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
fuzzybinary: It's the supported version of the JSON payload to intake, but it should likely be a constant in the companion object. Will fix.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdMappingFileUploadTask.kt
```diff
@@ -381,6 +388,11 @@ open class DdMappingFileUploadTask
     internal companion object {
         private const val INDENT = 4
 
+        internal const val TYPE_JVM_MAPPING_FILE = "jvm_mapping_file"
```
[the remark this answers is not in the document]
fuzzybinary: Yes. They happen to be the same string but they represent different things. One is the type of file (a value) and the other is the key in the form for the file.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTask.kt
fuzzybinary: I think so yes, but I'm a bit out of my element with Gradle on this front.

From what I can tell I can't inject the completed files collection, and I can only create one from the ObjectFactory, and I don't want it to be null.  So I create it during initialization from the injected ObjectFactory, then fill it as part of Register.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
fuzzybinary: Probably. This was straight copied from DdMappingFileUploadTask, but maybe I can revisit here and make it cleaner.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTask.kt
mariusc83: Yes you cannot inject but maybe you can put it as `lateinit` and not null instead without having to initialize it.

mariusc83: also I think what you nee there is more this lazy initialization of properties:
https://docs.gradle.org/current/userguide/lazy_configuration.html

alai97 (dismissed):
Looks good for docs!

mariusc83 (approved):

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTaskTest.kt
```diff
@@ -0,0 +1,666 @@
...
+import java.io.File
+import java.util.UUID
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+internal class DdSymbolFileUploadTaskTest {
```
0xnm, line 48: It is also worth adding some functional tests (see `DdAndroidGradlePluginFunctionalTest`), especially tests checking that if we call upload mapping task and also upload NDK symbol task for the same build, they will use the same build ID.

samples/ndk/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+        applicationId "com.datadog.example.ndk"
+        minSdkVersion AndroidConfig.MIN_SDK
+        targetSdkVersion AndroidConfig.TARGET_SDK
+        versionCode 1
+        versionName "1.0"
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        externalNativeBuild {
+            cmake {
+                cppFlags ''
```
0xnm, line 20: do we need to specify it explicitly?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/Uploader.kt
```diff
@@ -11,11 +11,23 @@ import com.datadog.gradle.plugin.RepositoryInfo
...
 
 internal interface Uploader {
+    // region File Info
+
+    data class UploadFileInfo(
+        val fileKey: String,
+        val file: File,
+        val encoding: String,
+        val fileType: String,
+        val fileName: String,
```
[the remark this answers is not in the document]
fuzzybinary: 👍 Good call, I'll add the documentation, where I'll also clear up that the fileName is the name of the file sent to the server, which can be different from the name of the file locally.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,159 @@
...
+            return project.tasks.register(
+                TASK_NAME + variant.name.capitalize(),
+                DdSymbolFileUploadTask::class.java
+            ) { task ->
+
+                nativeBuildProviders.forEach { buildTask ->
+                    val searchFiles = buildTask.flatMap {
+                        val soFolder =
+                            ExternalNativeBuildTask::class.memberProperties.find { it.name == "soFolder" }
+                                ?.get(it)!!
```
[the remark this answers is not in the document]
fuzzybinary: Apparently yes? I have no idea why I used reflection here.

fuzzybinary:
Resolved most of the comments -- I still need to do the functional tests so I'll jump on that next.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -94,6 +94,11 @@ class DdAndroidGradlePlugin @Inject constructor(
         } else {
             LOGGER.info("Minifying disabled for variant ${variant.name}, no upload task created")
         }
+        configureSymbolUploadTask(
```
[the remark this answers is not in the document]
0xnm: Yes, I saw that. I mean maybe it makes sense for the `register` method to return non-null value always, and do the check for the native build providers before the `register` call, but then I guess it will be harder to unit-test it.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
```diff
@@ -0,0 +1,326 @@
...
+                        versionCode = versionCode,
+                        variant = variantName,
+                        buildId = buildId.get()
+                    ),
+                    repositories.firstOrNull(),
+                    !disableGzipOption.isPresent
+                )
+            } catch (e: Exception) {
+                LOGGER.error("Failed to upload ${mappingFile.fileType} file", e)
+                lastError = e
```
0xnm, line 182: maybe instead we can collect all of them, and if there is only one, re-throw it as-is, and if there are multiple, create a new exception and add all of them to it using https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#addSuppressed%28java.lang.Throwable%29? wdyt? this way we are not losing any.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,191 @@
...
+            "Uploads NDK symbol files to Datadog servers to perform native crash symbolication."
+    }
+
+    override fun getFilesList(): List<Uploader.UploadFileInfo> {
+        val files = mutableListOf<Uploader.UploadFileInfo>()
+
+        searchDirectories
+            .flatMap(this::findSoFiles)
+            .toSortedSet(compareBy { it.absolutePath })
+            .toSet()
```
0xnm, line 43: still the same comment I posted before https://github.com/DataDog/dd-sdk-android-gradle-plugin/pull/220#discussion_r1479404196, imo this can be simply `toSet` here.

```suggestion
            .toSet()
```

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,191 @@
...
+            val major = groups[0].toIntOrNull()
+            val minor = groups[1].toIntOrNull()
+            val patch = groups[2].substringBefore("-").toIntOrNull()
+            if (major == null || minor == null || patch == null) return false
+            return major >= 7 && minor >= 0 && patch >= 0
+        }
+
+        private fun fixNativeOutputPath(taskFolder: File): File {
+            return taskFolder.parentFile.parentFile.takeIf { it.parentFile.name == "cxx" }
+                ?: taskFolder.parentFile.parentFile.parentFile.takeIf { it.parentFile.name == "cxx" }
```
0xnm, line 120: How can we be sure that `parentFile` always non-null? I think this API is coming from Java, so Kotlin cannot validate if it is null or non-null, but essentially, to prevent NPE we need either do the check or `?.` calls after `parentFile`

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,191 @@
...
+                ?: taskFolder
+        }
+
+        private fun getSearchDir(buildTask: TaskProvider<ExternalNativeBuildTask>, propertyName: String, providerFactory: ProviderFactory): Provider<File> {
+            return buildTask.flatMap { task ->
+                val soFolder = ExternalNativeBuildTask::class.memberProperties.find { it.name == propertyName }?.get(task)
```
0xnm, line 126: So reflection in the end? cannot we access property directly for above certain AGP versions?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,191 @@
...
+                when (soFolder) {
+                    is File -> providerFactory.provider { fixNativeOutputPath(soFolder) }
+                    is DirectoryProperty -> soFolder.map { fixNativeOutputPath(it.asFile) }
+                    else -> providerFactory.provider { File("") }
```
0xnm, line 130: I think changing method signature to `Provider<File?>` and returning `providerFactory.provider { null }` will be cleaner here.

Or maybe it is better even to throw and exception, we don't have anything to upload anyway, and it is during the task execution stage, not task configuration stage, so for the end-user it will be fine to see task failure (and we will know that we need a change here).

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,191 @@
...
+            return major >= 7 && minor >= 0 && patch >= 0
+        }
+
+        private fun fixNativeOutputPath(taskFolder: File): File {
+            return taskFolder.parentFile.parentFile.takeIf { it.parentFile.name == "cxx" }
+                ?: taskFolder.parentFile.parentFile.parentFile.takeIf { it.parentFile.name == "cxx" }
+                ?: taskFolder
+        }
+
+        private fun getSearchDir(buildTask: TaskProvider<ExternalNativeBuildTask>, propertyName: String, providerFactory: ProviderFactory): Provider<File> {
```
0xnm, line 124: seems too long, needs to be wrapped

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,191 @@
...
+            variant: ApplicationVariant,
+            buildIdTask: TaskProvider<GenerateBuildIdTask>,
+            providerFactory: ProviderFactory,
+            apiKey: ApiKey,
+            extensionConfiguration: DdExtensionConfiguration,
+            repositoryDetector: RepositoryDetector
+        ): TaskProvider<DdNdkSymbolFileUploadTask>? {
+            val nativeBuildProviders = variant.externalNativeBuildProviders
+            if (nativeBuildProviders.isEmpty()) {
+                LOGGER.info("No native build tasks found for variant ${variant.name}, skipping symbol file upload.")
```
0xnm, line 147:
```suggestion
                LOGGER.info("No native build tasks found for variant ${variant.name}, skipping NDK symbol file upload.")
```

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,191 @@
...
+                DdNdkSymbolFileUploadTask::class.java
+            ) { task ->
+                val roots = mutableListOf<File>()
+                variant.sourceSets.forEach { it ->
+                    roots.addAll(it.javaDirectories)
+                    if (isAgp7OrAbove()) {
+                        roots.addAll(it.kotlinDirectories)
+                    }
+                }
+                task.sourceSetRoots = roots
```
0xnm, line 162: That is actually the thing we **need** to improve somehow (no idea how for now), because we will upload repository info several times (and this repo info may be big, big projects may have thousands of files). Ideally we need to make an upload only once, but with that approach we will do upload for JVM mapping file, plus for each `.so` file we found.

I would say we shouldn't upload repo info for NDK symbol files, because this info will be uploaded with JVM mapping file, but then we may have a situation when app has NDK symbols, but no obfuscation (hence no mapping file).

I guess it will be too complicated to address this in this PR, so better to create a task for the future (maybe even some backend changes will be needed).

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,191 @@
...
+            searchDirectory.walkTopDown()
+                .filter { it.extension == "so" }
+                .toSet()
+        } else {
+            return emptySet()
+        }
+    }
+
+    companion object {
+        internal const val TASK_NAME = "ddUploadNdkSymbolFiles"
```
0xnm, line 75: same comment as https://github.com/DataDog/dd-sdk-android-gradle-plugin/pull/220#discussion_r1479412323

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
```diff
@@ -65,7 +65,9 @@ internal class GitRepositoryDetector(
         sourceSetRoots: List<File>
     ): List<String> {
         val files = mutableListOf<String>()
+        LOGGER.info("Listing ${sourceSetRoots}")
         sourceSetRoots.forEach { sourceSetRoot ->
+            LOGGER.info("Listing ${sourceSetRoot.absolutePath}")
```
0xnm, line 70: seems excessive, the log entry above will already contain the item logged here.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -939,9 +945,127 @@ internal class DdAndroidGradlePluginFunctionalTest {
...
+
+        val result = gradleRunner { withArguments("--info", ":samples:app:assembleRelease") }
+            .build()
+
+        // Then
+        assertThat(result).hasNoNdkSymbolUploadTasks()
+    }
+
+    @Test
+    fun `M try to upload the symbol file W upload { using a fake API_KEY }`(forge: Forge) {
```
0xnm, line 967: We need also to add the test where we do JVM mapping upload and NDK symbols upload, verifying that **all** uploads have the same `build_id` and it matches the `build_id` in the APK/AAB. This is very important for the consistency, because all uploads for the same build (variant) should share the same `build_id`.

0xnm (commented):
great! I added some comments, plus there are some comments to check still from the previous review.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/GitRepositoryDetector.kt
fuzzybinary: Yeah, sorry, left in from debugging. Nice catch.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
```diff
@@ -0,0 +1,326 @@
...
+     * Source of the API key set: environment, gradle property, etc.
+     */
+    @get:Input
+    var apiKeySource: ApiKeySource = ApiKeySource.NONE
+
+    /**
+     * The repository detector to use for detecting the repository information.
+     */
+    @get:Internal
+    var repositoryDetector: RepositoryDetector? = null
```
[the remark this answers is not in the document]
fuzzybinary: Would need to go through the service provider since we're using `register` and not constructing the object directly. I meant to look back into it and forgot. I'll see what I can see.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
fuzzybinary: Yeah that sounds good. I'll make that change.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
fuzzybinary: I'll give it a try and see what I can do.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
fuzzybinary: I thought I couldn't change the return to `Provider<File?>` for some reason... 😅

I think throwing the exception is probably not the best solution because we can have multiple .so files, and I don't think we want to fail all the others if just one fails.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
0xnm: Ideally we need to scan Git repo only once as well, because scan is expensive, it involves `git` CLI call. But yeah, let's address it in the future tickets.

fuzzybinary: Created as RUM-3171 and RUM-3172

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
fuzzybinary: I do get an issue returning null occasionally complaining that the property has no value, but in functional testing this has gone away now that everything surrounding it is working....

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
fuzzybinary: I'm removing this. It doesn't seem necessary. I'm sure we'll learn why it's there in the future.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,159 @@
...
+    override fun getFilesList(): List<Uploader.UploadFileInfo> {
+        val files = mutableListOf<Uploader.UploadFileInfo>()
+
+        searchDirectories
+            .flatMap(this::findSoFiles)
+            .toSortedSet(compareBy { it.absolutePath })
+            .toSet()
+            .forEach { file ->
+                val arch = file.parentFile.name
+                require(SUPPORTED_ARCHS.contains(arch))
```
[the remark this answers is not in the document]
fuzzybinary: It was, but now I'm reconsidering.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,159 @@
...
+
+        @Suppress("StringLiteralDuplication")
+        private fun resolveDatadogRepositoryFile(target: Project): File {
+            val outputsDir = File(target.buildDir, "outputs")
+            val reportsDir = File(outputsDir, "reports")
+            val datadogDir = File(reportsDir, "datadog")
+            return File(datadogDir, "repository.json")
+        }
+
+        internal fun findDatadogCiFile(projectDir: File): File? {
```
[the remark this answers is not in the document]
fuzzybinary: Was thinking about doing that as part of a larger refactor to have DdMappingFileUploadTask inherit from DdFileUploadTask. Basically adding them here for now then remove them when the two upload tasks share the code.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTask.kt
```diff
@@ -0,0 +1,159 @@
...
+            return emptySet()
+        }
+    }
+
+    companion object {
+        internal const val TASK_NAME = "ddUploadNdkSymbolFiles"
+        internal const val KEY_NDK_SYMBOL_FILE = "ndk_symbol_file"
+        internal const val TYPE_NDK_SYMBOL_FILE = "ndk_symbol_file"
+        internal const val ENCODING = "application/octet-stream"
+        internal val SUPPORTED_ARCHS = setOf("armeabi-v7a", "arm64-v8a", "x86", "x86_64")
```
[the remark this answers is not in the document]
fuzzybinary: https://developer.android.com/ndk/guides/abis

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTask.kt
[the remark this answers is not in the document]
fuzzybinary: Yeah, I think we should leave this in. If we're getting a weird ARCH we probably want to fast fail so we get a bug report.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
0xnm: Maybe then use `Provider#isPresent` before querying the value if using `Provider<File?>`. My concern here is that if in some future version type of `soFolder` is not `File` or `DirectoryProperty`, then we will silently miss it by using hacky `File("")` return value. We should notify about that.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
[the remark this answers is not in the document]
0xnm: You can still provide this value in the following [register](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskContainer.html#register-java.lang.String-java.lang.Class-java.lang.Object...-) method overload the same way it is done for `create`, so value is injected in the constructor and you don't have a nullable type here.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdSymbolFileUploadTask.kt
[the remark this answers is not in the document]
0xnm: As long as the exception is thrown during task execution phase (and it is), and not task configuration phase, I'm fine.

The difference is that task configuration phase is done with other tasks configuration. So if there is an error during the configuration phase, other plugins/tasks won't be configured (if I'm not wrong) as well, this will have big impact.

Task execution phase on another hand is independent (well, if it has no other tasks dependent), so error during the execution has much smaller impact.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -939,9 +944,199 @@ internal class DdAndroidGradlePluginFunctionalTest {
...
+                mappingUploadTaskName,
+                "--info",
+                "--stacktrace",
+                "-PDD_API_KEY=fakekey"
+            )
+        }
+            .buildAndFail()
+
+        // Then
+        val buildIdInOriginFile = testProjectDir.findBuildIdInOriginFile(variant)
```
0xnm, line 1105: paranoidal check: let's check also that this build ID exists in the apk/aab generated as well, after you executed upload tasks (to make sure that upload tasks don't re-trigger APK/AAB generation once again).

0xnm (approved):
looks great, nice work! I added few comments, but they are not blocking.

mariusc83 (dismissed):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
fuzzybinary: I changed it to return null, which will fail properly.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
```diff
@@ -0,0 +1,327 @@
...
+            MISSING_BUILD_ID_ERROR
+        }
+
+        val mappingFiles = getFilesList()
+        if (mappingFiles.isEmpty()) return
+
+        val repositories = repositoryDetector?.detectRepositories(
+            sourceSetRoots,
+            remoteRepositoryUrl
+        ).orEmpty()
```
0xnm, line 148: now once `repositoryDetector` is not null, this can be:

```suggestion
        val repositories = repositoryDetector.detectRepositories(
            sourceSetRoots,
            remoteRepositoryUrl
        )
```

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdFileUploadTask.kt
```diff
@@ -0,0 +1,327 @@
...
+            } catch (e: Exception) {
+                caughtErrors.add(e)
+            }
+        }
+        // If any errors occurred, throw them as a single exception
+        if (caughtErrors.isNotEmpty()) {
+            if (caughtErrors.count() == 1) {
+                throw caughtErrors.first()
+            } else {
+                val consolidatedError = Exception("Multiple errors occurred during upload")
```
0xnm, line 183: Didn't notice it on the previous run, sorry. It is better to use `RuntimeException` here, normally nobody instantiates parent class `Exception` itself.

```suggestion
                val consolidatedError = RuntimeException("Multiple errors occurred during upload")
```

0xnm (dismissed):

fuzzybinary:
@0xnm Fixed and amended the previous commit

0xnm (approved):

PR #221, opened 2024-02-09
chore: Update CMAKE to 3.22.1 for gitlab builds.

fuzzybinary:
### What does this PR do?

Some plugin builds are failing because we're missing CMAKE 3.22.1. Update the Dockerfile so that CMAKE 3.22.1 is always installed.


Dockerfile.gitlab
```diff
@@ -29,7 +29,7 @@ ENV ANDROID_COMPILE_SDK 34
 ENV ANDROID_BUILD_TOOLS 34.0.0
 ENV ANDROID_SDK_TOOLS 10406996
 ENV NDK_VERSION 21.3.6528147
-ENV CMAKE_VERSION 3.10.2.4988404
+ENV CMAKE_VERSION 3.22.1
```
0xnm, line 32: you probably want also to bump image version in `.gitlab-ci.yml`, so that it is deployed and used.

0xnm (approved):

Dockerfile.gitlab
fuzzybinary: Is that just changing [this line](https://github.com/DataDog/dd-sdk-android-gradle-plugin/blob/5895876b505bcfdc65a040b5fef036f696e539a5/.gitlab-ci.yml#L7) from 6 to 7? Do I have to do anything else?

0xnm: yes, you need to change this line and then run image generation job (`ci-image`) manually

0xnm (approved):

PR #249, opened 2024-04-19
fix: Standardize architectures

fuzzybinary:
### What does this PR do?

The symbolication service currently understands `arm`, `arm64`, `x86` and `x64` as valid symbol file architectures. Instead of adding the Android ABIs, we map them to those architectures instead.

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
@@ -42,7 +42,8 @@ internal abstract class DdNdkSymbolFileUploadTask @Inject constructor(
             .toSet()
             .forEach { file ->
                 val arch = file.parentFile.name
-                require(SUPPORTED_ARCHS.contains(arch))
+                val archMapping = SUPPORTED_ARCHS.firstOrNull { it.arch == arch }
+                require(archMapping != null)
                 files.add(
                     Uploader.UploadFileInfo(
                         KEY_NDK_SYMBOL_FILE,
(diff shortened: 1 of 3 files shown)


xgouchet (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
```diff
@@ -70,12 +71,23 @@ internal abstract class DdNdkSymbolFileUploadTask @Inject constructor(
...
         internal const val KEY_NDK_SYMBOL_FILE = "ndk_symbol_file"
         internal const val TYPE_NDK_SYMBOL_FILE = "ndk_symbol_file"
         internal const val ENCODING = "application/octet-stream"
-        internal val SUPPORTED_ARCHS = setOf("armeabi-v7a", "arm64-v8a", "x86", "x86_64")
+        internal val SUPPORTED_ARCHS = setOf(
+            SupportedArchitectureMapping("armeabi-v7a", "arm"),
+            SupportedArchitectureMapping("arm64-v8a", "arm64"),
+            SupportedArchitectureMapping("x86", "x86"),
+            SupportedArchitectureMapping("x86_64", "x64")
+        )
```
0xnm, line 90: Sounds like could be simply `mapOf<String, String>`, without introducing data class. But not a big deal.

0xnm (approved):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdNdkSymbolFileUploadTask.kt
fuzzybinary: Yeah, I was curious about that. I could go either way but I figured the data class was more descriptive.

0xnm: just lookup is now linear instead of O(1), but it doesn't really matter here

PR #309, opened 2025-01-28
Log message if there are no files to upload

0xnm:
### What does this PR do?

Small improvement to log the message if there are no files to upload in the upload task.

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/FileUploadTask.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/FileUploadTask.kt
@@ -151,7 +151,10 @@ abstract class FileUploadTask @Inject constructor(
         }
 
         val mappingFiles = getFilesList()
-        if (mappingFiles.isEmpty()) return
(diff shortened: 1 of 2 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/FileUploadTask.kt
```diff
@@ -151,7 +151,10 @@ abstract class FileUploadTask @Inject constructor(
         }
 
         val mappingFiles = getFilesList()
-        if (mappingFiles.isEmpty()) return
+        if (mappingFiles.isEmpty()) {
+            LOGGER.info("No mapping files to upload.")
```
xgouchet, line 155: > [!NOTE]
> Should it be a warning instead of info ? By default, `info` level logs are not visible when running gradle. My guess is that running this task, mapping files are expected to exist, which could be a good enough reason to make the log a warning.

0xnm: Yes, fair point. Updated.

xgouchet (approved):

PR #317, opened 2025-02-25
Merge develop into feature/kotlin-compiler-plugin

ambushwork:
### What does this PR do?
Merge develop into feature/kotlin-compiler-plugin

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -90,6 +90,7 @@ test:plugin:
   stage: test
   timeout: 1h
   script:
+    - aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android-gradle-plugin.gradle-properties --with-decryption --query "Parameter.Value" --out text >> ./gradle.properties
     - git fetch --depth=1 origin main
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -90,6 +90,7 @@ test:plugin:
   stage: test
   timeout: 1h
   script:
+    - aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android-gradle-plugin.gradle-properties --with-decryption --query "Parameter.Value" --out text >> ./gradle.properties
```
0xnm, line 93: I read that it is coming from https://github.com/DataDog/dd-sdk-android-gradle-plugin/pull/316, but I'm not sure why do we need to store Gradle Properties in SSM and not simply use run flags for CI (like we do with `--no-daemon` already).

You can have all flags defined in the env variable for this job and simply read it from there when running Gradle task.

ambushwork: develop branch is updated

0xnm (approved):

PR #322, opened 2025-02-28
RUM-8838: Register Kotlin compiler plugin to Kotlin compile options

ambushwork:
### What does this PR do?

Prior to this PR, in the sample app we need to use `kotlinCompilerPluginClasspath` to make KCP work during the kotlin compilation.
```
dependencies{
      kotlinCompilerPluginClasspath("com.datadoghq:dd-sdk-android-gradle-plugin:$version")
}
```
The goal of this PR is that KCP can be automatically registered after client apply our AGP in the `plugins{}`block:
```
plugins{
     id("com.datadoghq.dd-sdk-android-gradle-plugin")
}
```
The idea of this PR is to apply compilerOptions programatically from our AGP class
```
android {
    ...
    kotlin {
      compilerOptions {
          freeCompilerArgs.add(
              "-Xplugin=/path/to/kotlin-compiler-plugin.jar"
          )
    }
}
```
The step to achieve it:
* find plugin "org.jetbrains.kotlin.android" and configure insdie.
* Retrieve the "dd-sdk-android-gradle-plugin" version declared in root gradle classpath.
* Create a custom dependency configuration called "ddCompilerPluginClassPath" dedicated for KCP dependency
* Add "dd-sdk-android-gradle-plugin" dependency with "ddCompilerPluginClassPath" into `dependencies` block
* Retrieve the jar file path of the plugin from this dependency
* Add it into `freeCompilerArgs` as "-Xplugin=/path/to/kotlin-compiler-plugin.jar"

### Motivation

RUM-8838

### Additional Notes

RUM-8838

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
@@ -27,7 +27,9 @@ import org.gradle.api.provider.ProviderFactory
 import org.gradle.api.tasks.TaskContainer
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.process.ExecOperations
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.io.File
+import java.net.URISyntaxException
 import javax.inject.Inject
 import kotlin.io.path.Path
 
@@ -46,6 +48,10 @@ class DdAndroidGradlePlugin @Inject constructor(
     override fun apply(target: Project) {
         val extension = target.extensions.create(EXT_NAME, DdExtension::class.java)
         val apiKey = resolveApiKey(target)
+
+        target.pluginManager.withPlugin("org.jetbrains.kotlin.android") {
+            configureKotlinCompilerPlugin(target)
+        }
 
         // need to use withPlugin instead of afterEvaluate, because otherwise generated assets
         // folder with buildId is not picked by AGP by some reason
@@ -365,6 +371,40 @@ class DdAndroidGradlePlugin @Inject constructor(
         return configuration
     }
 
+    @Suppress("ReturnCount")
+    private fun configureKotlinCompilerPlugin(project: Project) {
+        val pluginJarPath = try {
+            val codeSource = this::class.java.protectionDomain?.codeSource
+            codeSource?.location?.toURI()?.path
+        } catch (e: URISyntaxException) {
+            LOGGER.error(
+                "Can not parse Datadog Gradle Plugin path because the URI is not correctly formatted.",
+                e
+            )
+            return
+        } catch (e: SecurityException) {
+            LOGGER.error(
+                "Failed to access Datadog Gradle Plugin protection domain due to insufficient permissions.",
+                e
+            )
+            return
+        }
+
+        if (pluginJarPath == null) {
+            LOGGER.warn(
+                "$DD_PLUGIN_MAVEN_COORDINATES not found in classpath, " +
+                    "Skipping Kotlin Compiler Plugin configuration."
+            )
+            return
+        }
+
+        project.tasks.withType(KotlinCompile::class.java).configureEach { task ->
+            task.kotlinOptions.freeCompilerArgs += listOf(
+                "-Xplugin=$pluginJarPath"
+            )
+        }
+    }
+
     private fun Project.stringProperty(propertyName: String): String? {
         return findProperty(propertyName)?.toString()
     }
@@ -390,6 +430,8 @@ class DdAndroidGradlePlugin @Inject constructor(
     companion object {
 
         private const val DD_FORCE_LEGACY_VARIANT_API = "dd-force-legacy-variant-api"
+
+        private const val DD_PLUGIN_MAVEN_COORDINATES = "com.datadoghq:dd-sdk-android-gradle-plugin"
 
         internal const val DD_API_KEY = "DD_API_KEY"
 
--- a/dd-sdk-android-gradle-plugin/build.gradle.kts
+++ b/dd-sdk-android-gradle-plugin/build.gradle.kts
@@ -49,6 +49,7 @@ dependencies {
     testImplementation(libs.kotlinPluginGradle)
     testImplementation(libs.kotlinCompilerEmbeddable)
 
+    compileOnly(libs.kotlinPluginGradle)
     compileOnly(libs.kotlinCompilerEmbeddable)
     compileOnly(libs.autoServiceAnnotation)
     kapt(libs.autoService)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -365,6 +370,43 @@ class DdAndroidGradlePlugin @Inject constructor(
...
+    private fun configureKotlinCompilerPlugin(project: Project) {
+        // Creates a new Gradle configuration to resolve the dependency differently.
+        val compilerPluginConfig = project.configurations.create(DD_COMPILER_PLUGIN_CONFIGURATION) {
+            it.isVisible = false
+            it.isCanBeResolved = true
+            it.isCanBeConsumed = false
+        }
+        val pluginVersion = resolvePluginVersion(project)
+        if (pluginVersion == null) {
+            LOGGER.warn("$AGP_FULL_PATH' not found in classpath, skip kotlin compiler plugin configuration.")
```
0xnm, line 382:
```suggestion
            LOGGER.warn("$AGP_FULL_PATH not found in classpath, skipping Kotlin Compiler Plugin configuration.")
```

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -365,6 +370,43 @@ class DdAndroidGradlePlugin @Inject constructor(
...
+            return
+        }
+
+        project.dependencies.add(compilerPluginConfig.name, "$AGP_FULL_PATH:$pluginVersion")
+
+        // Transitive dependencies are included in `compilerPluginConfig.files`, so we need to filter it by name.
+        val pluginJarPath = compilerPluginConfig.files.first {
```
0xnm, line 389: maybe use `firstOrNull`? even though I cannot think about the situation when configuration won't be resolved, it is better to be on the safe side.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -365,6 +370,43 @@ class DdAndroidGradlePlugin @Inject constructor(
...
+    }
+
+    private fun resolvePluginVersion(project: Project): String? {
+        // Gradle plugin is added to classpath in the gradle of project level.
+        val classpathConfig = project.rootProject.buildscript.configurations.findByName("classpath")
+
+        return classpathConfig
+            ?.resolvedConfiguration
+            ?.firstLevelModuleDependencies
+            ?.find { it.moduleGroup == AGP_GROUP_ID && it.moduleName == AGP_MODULE_ID }
```
0xnm, line 406: if plugin already makes a part of built-in configuration, why cannot we just traverse this configuration looking for the `jar` and need to create a fake configuration for the file resolution?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -365,6 +370,43 @@ class DdAndroidGradlePlugin @Inject constructor(
...
+        project.tasks.withType(KotlinCompile::class.java).configureEach { task ->
+            task.kotlinOptions.freeCompilerArgs += listOf(
+                "-Xplugin=$pluginJarPath"
+            )
+        }
+    }
+
+    private fun resolvePluginVersion(project: Project): String? {
+        // Gradle plugin is added to classpath in the gradle of project level.
+        val classpathConfig = project.rootProject.buildscript.configurations.findByName("classpath")
```
0xnm, line 401: is this a documented behavior?

0xnm (approved):
looks good! I've added some questions

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: updated

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: removed

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: indeed, actually we can have access to gradle plugin dependency, so now the plugin path changes from maven cache to gradle cache

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: removed

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -365,6 +370,21 @@ class DdAndroidGradlePlugin @Inject constructor(
         return configuration
     }
 
+    private fun configureKotlinCompilerPlugin(project: Project) {
+        val codeSource = this::class.java.protectionDomain?.codeSource
```
0xnm, line 374: I'm not sure if we should use this approach in production code, because of the security policies which may be enforced here (see https://docs.oracle.com/javase/8/docs/api/java/security/ProtectionDomain.html), but let's see how it goes.

It is a bit different approach from using resolved configuration, because former is relying on class loading mechanism, the latter is relying on simply download mechanism (which has no security policies, I believe).

If we are going to use this one, then we need to catch/handle all the exceptions which may happen till the final location resolution, because if any happens, then the whole build configuration process will be stopped and the only way to overcome this will be to remove our plugin completely. So if something happens here we want configuration process to continue.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -365,6 +370,21 @@ class DdAndroidGradlePlugin @Inject constructor(
...
+        val pluginJarPath = codeSource?.location?.toURI()?.path
```
0xnm, line 375: warning: while this works for now, because all the dependencies of the plugin (or, to be precise of Kotlin Compiler Plugin we are going to add) are provided and will be there, so that we may only pass our own JAR, in general what we should pass to `-Xplugin` should contain all JARs needed to be in the classpath to make the certain functionality work.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -365,6 +370,21 @@ class DdAndroidGradlePlugin @Inject constructor(
...
+        if (pluginJarPath == null) {
+            LOGGER.warn("$AGP_FULL_PATH' not found in classpath, Skipping Kotlin Compiler Plugin configuration.")
```
0xnm, line 377:
```suggestion
            LOGGER.warn("$AGP_FULL_PATH not found in classpath, Skipping Kotlin Compiler Plugin configuration.")
```

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -391,6 +411,8 @@ class DdAndroidGradlePlugin @Inject constructor(
 
         private const val DD_FORCE_LEGACY_VARIANT_API = "dd-force-legacy-variant-api"
 
+        private const val AGP_FULL_PATH = "com.datadoghq:dd-sdk-android-gradle-plugin"
```
0xnm, line 414: probably we should call it `DD_PLUGIN_MAVEN_COORDINATES`? there is nothing from AGP here.

0xnm (changes requested):

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: you mean we need also to pass transitive dependencies jar into the classpath as well? currently we only have a few no-transitive dependencies for KCP, and I currently don't see why we need to have some transitive ones in the future.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: updated

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: updated

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: Do you see any potential case that can cause the security policy issue here?

Anyway I add try catch block to catch `SecurityException` & `URISyntaxException` here, they seem to be the only possible exceptions here.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
0xnm: Yes, I mean if we will have, we will have to pass them as well. But yeah, indeed we have none.

But we will catch any issues around that with instrumented tests anyway.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
0xnm: I'm not sure in which case it can be enforced, maybe in some server env only though, or maybe in the future Java versions as a part of security hardening. But handling all the exception is a must here imo, so that build configuration can continue if something goes wrong here.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -365,6 +371,34 @@ class DdAndroidGradlePlugin @Inject constructor(
...
+    private fun configureKotlinCompilerPlugin(project: Project) {
+        val pluginJarPath = try {
+            val codeSource = this::class.java.protectionDomain?.codeSource
+            codeSource?.location?.toURI()?.path
+        } catch (e: URISyntaxException) {
+            LOGGER.error(e.message, e)
+            return
+        } catch (e: SecurityException) {
+            LOGGER.error(e.message, e)
+            return
```
0xnm, line 384: Maybe we should give our own message here? Like to explain better and also it will be easier to spot our thing in the customer logs.

0xnm (dismissed):
Added one suggestion, but overall lgtm.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: Good idea, updated

0xnm (approved):

ambushwork:
rebased to solve the conflicts

0xnm (approved):

PR #323, opened 2025-03-04
RUM-8876: Add Kotlin compilation unit test for modifier injection

ambushwork:
### What does this PR do?

In this PR, [Kotlin compile testing](https://github.com/tschuchortdev/kotlin-compile-testing) library is introduced to test the compilation result when datadog kotlin compiler plugin is applied.

The idea is to have a stub `DatadogModifier`, this stub one has the same path and function signature so that our KCP will be able to find it and inject it. 
Also for each case we have an example class with `@Composable` to verify that the injection will work in different use cases:
* No modifier is provided
* Default `Modifier` is provided
* Custom modifier is provided.

#### Testing Approach
A `TestCallbackContainer` is introduced to store the mock callback passed from the test environment. The verification process consists of the following steps:

* Load the `TestCallbackContainer` object using the compilation result’s class loader.
* Register the mock callback into `TestCallbackContainer` via reflection in the unit test function.
* Invoke the mock callback from the stub `DatadogModifier`.
* If the injection is successful, the composable function execution will trigger the mock callback.
* The unit test verifies whether the mock callback was invoked, confirming successful modifier injection.

### Motivation

RUM-8876

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -11,6 +11,10 @@ import,org.jetbrains,Apache-2.0,Copyright 2010-2019 JetBrains s.r.o. and Kotlin 
 import,org.jetbrains.kotlin,Apache-2.0,Copyright 2010-2019 JetBrains s.r.o. and Kotlin Programming Language contributors
 import,com.google.auto.service,Apache-2.0,Copyright 2013 Google LLC
 import,org.json,JSON,Copyright (c) 2002 JSON.org
+import(test),androidx.compose,Apache-2.0,Copyright 2019 The Android Open Source Project
+import(test),androidx.compose.ui,Apache-2.0,Copyright 2019 The Android Open Source Project
+import(test),com.android.tools.build,Apache-2.0,Copyright (C) 2013 The Android Open Source Project
+import(test),com.github.tschuchortdev,MPL-2.0,Copyright (C) 2023 Thilo Schuchort
 import(test),com.github.xgouchet.Elmyr,MIT,Copyright 2017-2019 Xavier F. Gouchet
 import(test),net.wuerl.kotlin,Apache-2.0,Copyright 2016 Andreas Würl
 import(test),org.assertj,Apache-2.0,Copyright 2012-2019 the original author or authors
--- a/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/KotlinCompilerTest.kt
+++ b/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/KotlinCompilerTest.kt
@@ -1,0 +1,341 @@
+package com.datadog.gradle.plugin
+
+import com.datadog.gradle.plugin.kcp.DatadogPluginRegistrar
+import com.tschuchort.compiletesting.KotlinCompilation
+import com.tschuchort.compiletesting.SourceFile
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.assertj.core.api.Assertions.assertThat
+import org.intellij.lang.annotations.Language
+import org.jetbrains.kotlin.compiler.plugin.ExperimentalCompilerApi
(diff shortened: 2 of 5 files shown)


LICENSE-3rdparty.csv
```diff
@@ -19,4 +19,8 @@ import(test),org.junit.jupiter,EPL-2.0,Copyright 2015-2019 the original author o
 import(test),org.junit.platform,EPL-2.0,Copyright 2015-2019 the original author or authors
 import(test),org.junit.vintage,EPL-2.0,Copyright 2015-2019 the original author or authors
 import(test),org.mockito,MIT,Copyright (c) 2007 Mockito contributors
-import(test),org.mockito.kotlin,MIT,"Copyright (c) 2016 Niek Haarman, Copyright (c) 2007 Mockito contributors"
\ No newline at end of file
+import(test),org.mockito.kotlin,MIT,"Copyright (c) 2016 Niek Haarman, Copyright (c) 2007 Mockito contributors"
+import(test),com.github.tschuchortdev,MPL-2.0,Copyright (C) 2023 Thilo Schuchort
+import(test),com.android.tools.build,Apache-2.0,Copyright (C) 2013 The Android Open Source Project
+import(test),androidx.compose,Apache-2.0,Copyright 2019 The Android Open Source Project
+import(test),androidx.compose.ui,Apache-2.0,Copyright 2019 The Android Open Source Project
```
0xnm, line 26: these should go up, per alphabetical order

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/KotlinCompilerTest.kt
```diff
@@ -0,0 +1,340 @@
...
+@OptIn(ExperimentalCompilerApi::class)
+class KotlinCompilerTest {
+
+    private val datadogModifierSourceFileContent = SourceFile.kotlin(
+        DD_MODIFIER_CLASS_FILE_NAME,
+        DD_MODIFIER_SOURCE_FILE_CONTENT
+    )
+
+    @Mock
+    private lateinit var mockDDModifierCallback: () -> Unit
```
0xnm, line 37: just for the naming consistency

```suggestion
    private lateinit var mockDatadogModifierCallback: () -> Unit
```

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/KotlinCompilerTest.kt
```diff
@@ -0,0 +1,340 @@
...
+        val result = compileFile(noModifierTestCaseSource, false)
+        executeClassFile(
+            classLoader = result.classLoader,
+            className = "com.datadog.kcp.NoModifierTestCase",
+            methodName = "NoModifierTestCase",
+            methodArgs = listOf()
+        )
+
+        // Then
+        assert(result.exitCode == KotlinCompilation.ExitCode.OK)
```
0xnm, line 66: this won't have any effect unless JVM is started with `-ea` (enable assertions) flags. Probably you've meant to use `assertThat(...).isEqualTo(...)` here?

Same for other test cases.

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/KotlinCompilerTest.kt
```diff
@@ -0,0 +1,340 @@
...
+        verify(mockDDModifierCallback, times(0)).invoke()
```
0xnm, line 67:
```suggestion
        verify(mockDDModifierCallback, never()).invoke()
```

Alternatively, since if is a lambda, you can use `verifyNoInteractions(mockDDModifierCallback)`

0xnm (changes requested):

0xnm (approved):

PR #341, opened 2025-03-21
RUM-8953: Add Instrumented test for view trackings instrumentation

ambushwork:
### What does this PR do?

Add instrumented test for view trackings instrumentation, the test verifies two things:

* The instrumentation doesn't break client's code. (the original components should still be present on the screen after instrumentation)
* The function that we instrumented takes effect. (`NavigationViewTrackingEffect` registers listener inside the `navController`, so we use reflection to verify that navController's listener size)

### Motivation

RUM-8953

### DIFF
--- a/instrumented/src/androidTest/java/com/datadog/android/instrumented/NavigationTest.kt
+++ b/instrumented/src/androidTest/java/com/datadog/android/instrumented/NavigationTest.kt
@@ -1,0 +1,108 @@
+package com.datadog.android.instrumented
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.ComposeContentTestRule
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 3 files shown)


instrumented/src/androidTest/java/com/datadog/android/instrumented/NavigationTest.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private fun ComposeContentTestRule.setContentWithStubLifeCycle(
+        composable: @Composable () -> Unit
+    ): StubLifecycleOwner {
+        var stubLifecycleOwner: StubLifecycleOwner? = null
+        InstrumentationRegistry.getInstrumentation().runOnMainSync {
+            stubLifecycleOwner = StubLifecycleOwner()
+        }
+        this.setContent {
```
0xnm, line 84: minor: you can insert `checkNotNull(stubLifecycleOwner)` here and I hope it will trigger smart-cast to non-nullable type for the usages below, so that `!!` are not needed.

0xnm (dismissed):

instrumented/src/androidTest/java/com/datadog/android/instrumented/NavigationTest.kt
ambushwork: `checkNotNull` seems not working for `var`, because after checking it, it is still possible to modify the variable, so compiler may think it is not safe.

0xnm (dismissed):

0xnm (approved):

PR #345, opened 2025-03-26
Enhance compiler plugin binary backward compatibility

ambushwork:
### What does this PR do?

The Kotlin compiler plugin relies on `Kotlin-embeddable` for Intermediate Representation phase, 
when plugin is running in the host application during the compilation,  the Kotlin version from Android Gradle Tool will override the one in plugin during the runtime. 

For example, if the Android Gradle Tool version is "8.8.2", the kotlin "1.9.24" version from it will be actually used during the runtime of the plugin despite "2.0.20" is declared in the plugin project. And since it's a major version change, it leads some binary compatibility issue which makes build failed.

The main idea of this PR is to check the known broken API version, by checking function signature, visibility, etc.. to instantiate it by reflection if we find it's not compatible with the compilation version.

Here is a list of breaking APIs:
|Usage| Version 1.9 |Version 2.0 | 
|--| -- | -- | 
|IrDeclarationOrigin.FAKE_OVERRIDE| object of an interface | companion object property of an interface | 
|IrDeclarationOrigin.LOCAL_FUNCTION_FOR_LAMBDA| object of an interface | companion object property of an interface |
|IrDeclarationOrigin.DEFINED| object of an interface | companion object property of an interface |
|IrFunctionExpressionImpl| a class constructor | top level function|
| IrValueParameterSymbolImpl| class constructor with one argument| class constructor with several arguments | 
| IrSimpleFunctionSymbolImpl| class constructor with one argument| class constructor with several arguments | 

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/kcp/IrExt.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/kcp/IrExt.kt
@@ -6,13 +6,14 @@ import org.jetbrains.kotlin.descriptors.DescriptorVisibility
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.ir.UNDEFINED_OFFSET
 import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
+import org.jetbrains.kotlin.ir.declarations.IrDeclarationOriginImpl
 import org.jetbrains.kotlin.ir.declarations.IrDeclarationParent
-import org.jetbrains.kotlin.ir.declarations.IrFactory
 import org.jetbrains.kotlin.ir.declarations.IrSimpleFunction
 import org.jetbrains.kotlin.ir.declarations.impl.IrFactoryImpl
 import org.jetbrains.kotlin.ir.expressions.IrBody
 import org.jetbrains.kotlin.ir.expressions.IrFunctionExpression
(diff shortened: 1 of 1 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/kcp/IrExt.kt
```diff
@@ -116,3 +106,78 @@ internal fun irSimpleFunction(
...
+    }
+}
+
+private fun getCompatDefinedOrigin(): IrDeclarationOriginImpl {
+    return getCompatDeclarationOrigin("DEFINED") {
+        IrDeclarationOrigin.DEFINED
+    }
+}
+
+private fun getCompatDeclarationOrigin(
```
0xnm, line 128: is it possible to check what is the version of the Kotlin compiler we are running on? If yes, we can skip reflection-based calls for the versions 2.x+ and use reflection only for 1.x.

In the current approach we are using reflection all the time, if I'm not wrong?

ambushwork: I think it's possible to know it by calling `KotlinVersion.CURRENT` during the runtime (not tested though).


But here we are actually skipping the reflection in 2.x+, because only in lower version the `IrDeclarationOriginImpl` is an abstract class, we use that to differentiate the version, and for 2.x+, it will call `default` lambda which returns `IrDeclarationOrigin.DEFINED`

0xnm (approved):

PR #361, opened 2025-04-11
RUM-9278: Add precise error message for missing dependencies

ambushwork:
### What does this PR do?

Priror to this PR, only a general compilation error will be thrown when any dependency is missing, that is not precise enough for client to understand the issue.
In this PR, error message is thrown with exact dependency path, for example if `androidx.navigation:navigation-compose` is missing, client will see this:

Screenshot 2025-04-11 at 09 47 45

### Motivation

RUM-9278

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/kcp/MessageCollectorExt.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/kcp/MessageCollectorExt.kt
@@ -1,0 +1,13 @@
+package com.datadog.gradle.plugin.kcp
+
+import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
+import org.jetbrains.kotlin.cli.common.messages.MessageCollector
+
+internal fun MessageCollector.abortError() {
+    report(CompilerMessageSeverity.ERROR, ERROR_MISSING_DEP)
+}
+
+private const val ERROR_MISSING_DEP =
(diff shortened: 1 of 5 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/kcp/MessageCollectorExt.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
+import org.jetbrains.kotlin.cli.common.messages.MessageCollector
+
+internal fun MessageCollector.abortError() {
+    report(CompilerMessageSeverity.ERROR, ERROR_MISSING_DEP)
+}
+
+private const val ERROR_MISSING_DEP =
+    "The Datadog Kotlin Compiler Plugin failed to initialize references and will abort.\n" +
```
0xnm, line 11: suggestion: let's use `Kotlin Compiler Plugin` only internally and not expose this to the customer, because it is a bit generic imo

```suggestion
    "The Datadog Compose Instrumentation failed to initialize references and will abort.\n" +
```

0xnm (dismissed):
lgtm, added one suggestion

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/kcp/MessageCollectorExt.kt
ambushwork: updated

0xnm (approved):

PR #408, opened 2025-07-16
RUM-10823: Update Kotlin Compiler Plugin to Kotlin 2.2.0

ambushwork:
### What does this PR do?

Update Kotlin Compiler Plugin to Kotlin 2.2.0

### Motivation

RUM-10823

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -31,6 +31,7 @@ dependencies {
 
     // Dependencies used to configure the gradle plugins
     implementation(embeddedKotlin("gradle-plugin"))
+    implementation(libs.kotlinPluginGradle)
     implementation(libs.androidToolsPluginGradle)
     implementation(libs.versionsPluginGradle)
     implementation(libs.fuzzyWuzzy)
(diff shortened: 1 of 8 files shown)


ambushwork:
> Just to better understand. Will our plugin work for projects where kotlin version is:
> 
> 1. ">= 2.0 and < 2.2.0" ?
> 2. < 2.0 ?

@aleksandr-gringauz with this fix, the plugin will only work for the kotlin version above 2.2.0. The user who has the lower version needs to either stay with the old version of plugin or bump their kotlin version.

jonathanmos (approved):

0xnm:
> the plugin will only work for the kotlin version above 2.2.0

I'm not quite comfortable with this change, because it forces users who want to use Compose Auto-Instrumentation to stick with particular Kotlin version.

Did we consider having slight different approach: extract Compose Auto-Instrumentation into a dedicated module/artifact sliced by the Kotlin version and make the main plugin pull the right version by using detached configuration?

ambushwork:
> > the plugin will only work for the kotlin version above 2.2.0
> 
> I'm not quite comfortable with this change, because it forces users who want to use Compose Auto-Instrumentation to stick with particular Kotlin version.
> 
> Did we consider having slight different approach: extract Compose Auto-Instrumentation into a dedicated module/artifact sliced by the Kotlin version and make the main plugin pull the right version by using detached configuration?

This solution has been denied by myself and the PR is closed, please check the new solution ongoing: https://github.com/DataDog/dd-sdk-android-gradle-plugin/tree/feature/kotlin-source-sets

PR #420, opened 2025-08-19
RUM-11262: Upgrade Kotlin Version to 2.1.21 to remove metadata check error

ambushwork:
### Context
Prior to this PR, we use Kotlin 2.0.21 for the whole project, when compiling the `kotlin22` source set, the compiler will throw the error saying:
> Class 'org.jetbrains.kotlin.backend.common.extensions.FirIncompatiblePluginAPI' was compiled with an incompatible version of Kotlin. The actual metadata version is 2.2.0, but the compiler version 2.0.0 can read versions up to 2.1.

Then we have to add following argument to the compiler to let it pass:
```
tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
    kotlinOptions {
        freeCompilerArgs += "-Xskip-metadata-version-check"
    }
}
```

### What does this PR do?

To remove this, we need to upgrade Kotlin compiler version to 2.1.x.

### Motivation

RUM-11262

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -30,7 +30,7 @@ repositories {
 dependencies {
 
     // Dependencies used to configure the gradle plugins
-    implementation(embeddedKotlin("gradle-plugin"))
+    implementation(libs.kotlinPluginGradle)
     implementation(libs.androidToolsPluginGradle)
     implementation(libs.versionsPluginGradle)
     implementation(libs.fuzzyWuzzy)
--- a/dd-sdk-android-gradle-plugin/build.gradle.kts
+++ b/dd-sdk-android-gradle-plugin/build.gradle.kts
@@ -81,8 +81,8 @@ dependencies {
 
     // Compile-only dependencies
     compileOnly(libs.androidToolsPluginGradle) // for auto-wiring into Android projects
-    compileOnly(libs.kotlinPluginGradle)
     compileOnly(libs.kotlinCompilerEmbeddable)
+    compileOnly(libs.kotlinPluginGradle)
     compileOnly(kotlin20.output)
     compileOnly(kotlin21.output)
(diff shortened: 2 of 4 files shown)


buildSrc/build.gradle.kts
```diff
@@ -30,7 +30,7 @@ repositories {
 dependencies {
 
     // Dependencies used to configure the gradle plugins
-    implementation(embeddedKotlin("gradle-plugin"))
+    implementation(libs.kotlinPluginGradle)
```
ambushwork, line 33: Note: we must switch to the designated version of plugin gradle (2.1.21), otherwise the version embedded in AGP is always 2.0.21

dd-sdk-android-gradle-plugin/build.gradle.kts
```diff
@@ -81,8 +81,8 @@ dependencies {
 
     // Compile-only dependencies
     compileOnly(libs.androidToolsPluginGradle) // for auto-wiring into Android projects
-    compileOnly(libs.kotlinPluginGradle)
     compileOnly(libs.kotlinCompilerEmbeddable)
+    compileOnly(libs.kotlinPluginGradle)
```
ambushwork, line 85: There are few classes (only used for auto instrumentation, such as CommandLineProcessor) existing both in `kotlinPluginGradle` and `kotlinCompilerEmbeddable`, the compiler will resolve it by the order defined here. (And also weird enough, this conflict happens only with kotlin 2.1.x version, neither in 2.0.x nor 2.2.x)

In this case I need to reverse the order so that compiler take always the one from `kotlinCompilerEmbeddable` when there are some duplications. May be we need to consider something safer, but this is the easiest fix that I can find for now.

0xnm (approved):

Issue #206, opened 2023-11-14
uploadMappingRelease sometimes leaks http connection & task ends up failing with timeout error


jamesbluecrow:
**Describe what happened**
I often see errors when running `uploadMappingRelease` about a http connection being leaked. When that happens the task ends up failing after 5-8 minutes running with the log below:

```
A connection to https://sourcemap-intake.datadoghq.com/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);

> Task :app:uploadMappingRelease FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:uploadMappingRelease'.
> Unable to upload mapping file with tags `service:"android-client-app"`, `version:4.2.22`, `variant:` because of a request timeout; check your network connection

```

**Steps to reproduce the issue:**
No clear repro steps, it happens more often when I'm out and the internet connection is less reliable but it also happens sometimes at home with fast internet.

**Describe what you expected:**
I think that either the task should succeed or the task should fail faster if the connection fails but there should not be a leaked connection.

**Additional context**
 - Android Build Tools version: 8.1.2
 - Gradle version: 8.4
 - OS version: Mac OS
 - Plugin version: 1.12.0
 - Proguard configuration: 
 - 
```
# Twilio Programmable Voice
-keep class com.twilio.** { *; }
-keep class tvo.webrtc.** { *; }
-dontwarn tvo.webrtc.**
-keep class com.twilio.voice.** { *; }
-keepattributes InnerClasses
# needed with AGP 8.x
-dontwarn android.content.pm.PackageManager$ApplicationInfoFlags
-dontwarn android.content.pm.PackageManager$PackageInfoFlags

-keepnames class * implements android.os.Parcelable
-keepnames class * implements java.io.Serializable
```

0xnm:
Hello @jamesbluecrow! Thanks for reporting it. Can you please also share the full exception you have? You can run the task with `--stacktrace` flag to see it.

0xnm:
Closing this issue as stale, we didn't get any new reports of such behavior. Feel free to re-open it if needed.

PR #467, opened 2025-11-17
[OkHttpUploader] Fix response not being closed in handleResponse()

theobch:
### What does this PR do?

`ResponseBody#close`(which closes the `ResponseBody`) is called only in case of `statusCode` equals or superior to 
400 (last statement in `handleResponse`) not in the other cases. Ensure `ResponseBody#close` is being called at the end of `handleResponse`. 
Also avoid instantiating unnecessary `call` in `upload` in case of emulated environment.

### Motivation

This log which happens during our build:

> [Timestamp] | main | A connection to https://sourcemap-intake.datadoghq.com/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);

### Additional Notes

N/A

- [x] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc)

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
@@ -197,7 +197,7 @@ internal class OkHttpUploader : Uploader {
                 }
             }
         }
-        response.close()
+        response.body?.close()
     }
 
     private fun validateApiKey(site: DatadogSite, apiKey: String): Boolean? {


0xnm:
Hello @theobch!

Thanks for opening PR. The issue looks similar to the https://github.com/DataDog/dd-sdk-android-gradle-plugin/issues/206, do you get this warning often?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/OkHttpUploader.kt
```diff
@@ -198,6 +197,7 @@ internal class OkHttpUploader : Uploader {
                 }
             }
         }
+        response.close()
```
0xnm: Generally speaking, this can lead to the exception according to the description here https://github.com/square/okhttp/blob/4984568367caaf359b82c452bd28b5e192824d1c/okhttp/src/main/kotlin/okhttp3/Response.kt#L295-L303, so it is better to call `close()` on the body itself.

PR #478, opened 2026-01-05
RUM-13459: Support AGP 9.0

### LINKED ISSUE #477: AGP 9.0 support
**Is your feature request related to a problem? Please describe.**
AGP 9.0 is around the corner. I see that DD plugin is complaining that applied not in Android project. But it is not correct.

**Describe the solution you'd like**
Support AGP 9.

0xnm:
### What does this PR do?

This PR brings support of AGP 9.0 and fixes #477.

The full list of changes coming with AGP 9.0 can be found here https://developer.android.com/build/releases/agp-preview.

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
@@ -6,6 +6,7 @@
 
 package com.datadog.gradle.plugin
 
+import com.android.build.api.dsl.ApplicationExtension
 import com.android.build.api.variant.ApplicationAndroidComponentsExtension
 import com.android.build.gradle.AppExtension
 import com.datadog.gradle.plugin.internal.ApiKey
@@ -67,7 +68,7 @@ class DdAndroidGradlePlugin @Inject constructor(
                     )
                 }
             } else {
-                val androidExtension = target.androidApplicationExtension ?: return@withPlugin
+                val androidExtension = target.legacyAndroidApplicationExtension ?: return@withPlugin
                 androidExtension.applicationVariants.all { variant ->
                     if (extension.enabled) {
                         configureTasksForVariant(
@@ -83,12 +84,19 @@ class DdAndroidGradlePlugin @Inject constructor(
 
         target.afterEvaluate {
             var isKcpEnabled = false
-            target.pluginManager.withPlugin("org.jetbrains.kotlin.android") {
+            if (CurrentAgpVersion.IMPLEMENTS_BUILT_IN_KOTLIN) {
                 isKcpEnabled = configureKotlinCompilerPlugin(target, extension)
+            } else {
+                target.pluginManager.withPlugin("org.jetbrains.kotlin.android") {
+                    isKcpEnabled = configureKotlinCompilerPlugin(target, extension)
+                }
             }
-            val androidExtension = target.androidApplicationExtension
-            if (androidExtension == null && !isKcpEnabled) {
-                LOGGER.error(ERROR_NOT_ANDROID)
+            if (!target.isAndroidAppPluginApplied) {
+                if (extension.composeInstrumentation != InstrumentationMode.DISABLE) {
+                    if (!isKcpEnabled) LOGGER.error(ERROR_COMPOSE_INSTRUMENTATION_NOT_APPLIED)
+                } else {
+                    LOGGER.error(ERROR_NOT_ANDROID)
+                }
             } else if (!extension.enabled) {
                 LOGGER.info(MSG_PLUGIN_DISABLED)
             }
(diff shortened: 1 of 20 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -83,12 +84,19 @@ class DdAndroidGradlePlugin @Inject constructor(
 
         target.afterEvaluate {
             var isKcpEnabled = false
-            target.pluginManager.withPlugin("org.jetbrains.kotlin.android") {
+            if (CurrentAgpVersion.IMPLEMENTS_BUILT_IN_KOTLIN) {
```
ambushwork: AFAIK Built in kotlin can be [opt-out](https://developer.android.com/build/migrate-to-built-in-kotlin#opt-out-of-built-in-kotlin), wondering in this case if just checking the version of AGP is fine enough?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -83,12 +84,19 @@ class DdAndroidGradlePlugin @Inject constructor(
...
+                target.pluginManager.withPlugin("org.jetbrains.kotlin.android") {
+                    isKcpEnabled = configureKotlinCompilerPlugin(target, extension)
+                }
             }
-            val androidExtension = target.androidApplicationExtension
-            if (androidExtension == null && !isKcpEnabled) {
-                LOGGER.error(ERROR_NOT_ANDROID)
+            if (!target.isAndroidAppPluginApplied) {
+                if (extension.composeInstrumentation != InstrumentationMode.DISABLE) {
+                    if (!isKcpEnabled) LOGGER.error(ERROR_JETPACK_INSTRUMENTATION_NOT_ENABLED)
```
ambushwork: I think the logic has changed here, now it's saying when the module is not Android **Application** project but KCP enabled, this error will be thrown. However KCP should be  able to be applied on the pure Android (but not Application) module, such as a Compose UI module.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
```diff
@@ -467,6 +493,9 @@ class DdAndroidGradlePlugin @Inject constructor(
 
         internal const val UPLOAD_TASK_NAME = "uploadMapping"
 
+        private const val ERROR_JETPACK_INSTRUMENTATION_NOT_ENABLED = "The dd-android-gradle-plugin has been applied" +
```
ambushwork: I think the error message and content are not corresponding.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
0xnm: It can be, but it is hard to imagine the project where Kotlin plugin is not applied, so with such fresh AGP version I don't think we will hit such situation.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
0xnm: Do you mean the variable name don't match the string value? Basically variable name says that we didn't manage to instrument the module (enable instrumentation), do you have a better name in mind?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: OK I was confused by the double check on L95&L96 and realised you change the return value in L425, then maybe the variable `isKcpEnabled` doesn't mean the same thing as before, it doesn't indicate the KCP is actually activated.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
ambushwork: well based on the discussion that we have in the other thread, this errors mean that the block of compose instrumentation is not successfully applied, (it doesn't represent the KCP is actually enabled or activated), so may be rename to `ERROR_COMPOSE_INSTRUMENTATION_NOT_APPLIED`?

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePlugin.kt
0xnm: It means that we managed to pass KCP extension options to the Kotlin plugin. The previous logic had a mistake by mixing the `composeInstrumentation` value to the same variable, because it was evaluating `isKcpEnabled` to false on the lib projects with `composeInstrumentation` disabled (which is a default value), giving a confusing warning in a totally correct situation.

jonathanmos (approved):

ambushwork (approved):

PR #522, opened 2026-03-16
RUM-15129: Fix Kotlin 2.3.20 breaking changes for Kotlin Compiler Plugin

ambushwork:
### What does this PR do?

This PR fixes the breaking change introduced in Kotlin compiler embeddable 2.3.20 .

#### What has changed?

in Kotlin [2.3.0](https://github.com/JetBrains/kotlin/blob/v2.3.0/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrDeclarationOrigin.kt#L59), `IrDeclarationOrigin.LOCAL_FUNCTION_FOR_LAMBDA` is declared as 
```
 val LOCAL_FUNCTION_FOR_LAMBDA by IrDeclarationOriginImpl
```

in Kotlin [2.3.20](https://github.com/JetBrains/kotlin/blob/v2.3.20/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/declarations/IrDeclarationOrigin.kt#L57), this field still exists but silently change the declaration under the hood, it becomes 
```
val LOCAL_FUNCTION_FOR_LAMBDA by IrDeclarationOriginImpl.Regular
```

#### How to fix

This can be covered by a tiny fix, so this fix is added in `kotlin22` module instead of creating a new module for this version.

Basically we check if `IrDeclarationOriginImpl.Regular` is introduced, if it is we use it to get `LOCAL_FUNCTION_FOR_LAMBDA` otherwise we fallback what we were using before.

### Motivation

RUM-15129

### DIFF
--- a/dd-sdk-android-gradle-plugin-kcp-kotlin22/src/main/kotlin/com/datadog/gradle/plugin/kcp/IrExt.kt
+++ b/dd-sdk-android-gradle-plugin-kcp-kotlin22/src/main/kotlin/com/datadog/gradle/plugin/kcp/IrExt.kt
@@ -13,6 +13,7 @@ import org.jetbrains.kotlin.descriptors.DescriptorVisibility
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.ir.UNDEFINED_OFFSET
 import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
+import org.jetbrains.kotlin.ir.declarations.IrDeclarationOriginImpl
 import org.jetbrains.kotlin.ir.declarations.IrDeclarationParent
 import org.jetbrains.kotlin.ir.declarations.IrParameterKind
 import org.jetbrains.kotlin.ir.declarations.IrSimpleFunction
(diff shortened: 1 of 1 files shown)


ambushwork:
@codex review

dd-sdk-android-gradle-plugin-kcp-kotlin22/src/main/kotlin/com/datadog/gradle/plugin/kcp/IrExt.kt
```diff
@@ -115,6 +119,46 @@ internal fun irSimpleFunction(
...
+    // changing the JVM return type of companion property getters (binary incompatibility).
+    // Use reflection to access the property by name when running on such versions.
+    val hasRegularSubtype = IrDeclarationOriginImpl::class.nestedClasses
+        .any { it.simpleName == "Regular" }
+    return if (hasRegularSubtype) {
+        IrDeclarationOrigin::class.companionObjectInstance?.let { companion ->
+            companion::class.memberProperties
+                .filterIsInstance<KProperty1<Any, *>>()
+                .first { it.name == name }
+                .get(companion) as IrDeclarationOrigin
```
0xnm: nit: `companion` is a reserved keyword in Kotlin, so maybe choose another name since it breaks a proper code highlight in GH UI at least.

0xnm (dismissed):

dd-sdk-android-gradle-plugin-kcp-kotlin22/src/main/kotlin/com/datadog/gradle/plugin/kcp/IrExt.kt
ambushwork: I believe only `companion object` makes it a keyword not `companion` it self, but updated to `instance` anyway in case of confusion.

0xnm: No, `companion object` is an expression, not just a keyword. `companion` as a keyword can be found here https://kotlinlang.org/docs/keyword-reference.html#modifier-keywords.

`companion` alone is enough to confuse Github code highlighting and highlight it as a part of the language syntax.

0xnm (approved):

Issue #516, opened 2026-03-08
Mapping file upload issue on CI/CD


electrolobzik:
Hi, I have an issue very similar to #462 

**Describe what happened**
```
> Task :app:compileQaReleaseJavaWithJavac
> Task :feature:more:lintVitalAnalyzeQaRelease
> Task :app:hiltAggregateDepsQaRelease
> Task :app:hiltJavaCompileQaRelease
> Task :app:processQaReleaseJavaRes
> Task :app:transformQaReleaseClassesWithAsm
> Task :app:mergeQaReleaseGeneratedProguardFiles
> Task :app:generateQaReleaseLintVitalReportModel
> Task :app:expandQaReleaseArtProfileWildcards
> Task :app:mergeQaReleaseJavaResource
> Task :app:lintVitalAnalyzeQaRelease

> Task :app:minifyQaReleaseWithR8
WARNING: /Users/builder/.gradle/caches/8.13/transforms/e1fb958f7d4c07b1299b3d8ef8c84b5c/transformed/play-services-location-21.3.0-runtime.jar: R8: The companion object Companion could not be found in class com.google.android.gms.********.location.zze

> Task :app:lintVitalReportQaRelease
> Task :app:lintVitalQaRelease
> Task :app:convertShrunkResourcesToBinaryQaRelease
> Task :app:compileQaReleaseArtProfile
> Task :app:optimizeQaReleaseResources
> Task :app:packageQaRelease
> Task :app:assembleQaRelease
> Task :app:createQaReleaseApkListingFileRedirect
> Task :app:uploadMappingQaRelease FAILED
> Task :app:uploadCrashlyticsMappingFileQaRelease

[Incubating] Problems report is available at: file:///Users/builder/clone/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:uploadMappingQaRelease'.
> Build ID is missing, you need to run upload task only after APK/AAB file is generated.
```

**Steps to reproduce the issue:**

I have the following gradle config:
```
afterEvaluate {
    tasks.matching { it.name.matches(Regex("(assemble|bundle).+Release")) }.configureEach {
        val variant = name.removePrefix("assemble").removePrefix("bundle")
        tasks.findByName("uploadMapping$variant")?.let { finalizedBy(it) }
    }
}
```

**Additional context**
 - Android Build Tools version: 8.13.2
 - Gradle version: 8.13
[...]

PR #530, opened 2026-03-31
RUM-15333: Fix missing build ID error in case of configuration cache warn mode

0xnm:
### What does this PR do?

This PR brings configuration cache support for the Upload Mapping task.

Before it was using [notCompatibleWithConfigurationCache](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api/-task/not-compatible-with-configuration-cache.html) API which has a specific behaviour in the configuration cache warn mode:

> The presence of incompatible tasks in the task graph will cause the configuration state to be discarded at the end of the build unless the global configuration-cache-problems option is set to warn, in which case the configuration state would still be cached in a best-effort manner as usual for the option.

That was causing `Build ID is missing ...` error.

Potentially it may address https://github.com/DataDog/dd-sdk-android-gradle-plugin/issues/516.

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTestBase.kt
+++ b/dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTestBase.kt
@@ -18,10 +18,8 @@ import fr.xgouchet.elmyr.annotation.StringForgery
 import fr.xgouchet.elmyr.junit5.ForgeConfiguration
 import fr.xgouchet.elmyr.junit5.ForgeExtension
 import org.gradle.api.Project
-import org.gradle.api.Transformer
 import org.gradle.api.file.RegularFile
 import org.gradle.api.provider.Provider
-import org.gradle.api.tasks.TaskProvider
 import org.gradle.testfixtures.ProjectBuilder
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.extension.ExtendWith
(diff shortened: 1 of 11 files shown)


ambushwork (dismissed):

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTestBase.kt
```diff
@@ -96,8 +94,8 @@ internal abstract class DdAndroidGradlePluginTestBase {
     protected fun mockBuildIdGenerationTask(buildId: String): TaskProvider<GenerateBuildIdTask> {
         return mock<TaskProvider<GenerateBuildIdTask>>().apply {
             whenever(
-                flatMap(any<Transformer<Provider<String>, GenerateBuildIdTask>>())
-            ) doReturn buildId.asProvider()
+                flatMap(any<Transformer<Provider<RegularFile>, GenerateBuildIdTask>>())
+            ) doReturn buildId.asBuildIdFileProvider()
```
hamorillo: ❓ _nit_: I would say we can safely remove the stub for flatMap, since we are no longer using that approach, right?

hamorillo (dismissed):
🚀 Nice!

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginTestBase.kt
0xnm: nice catch! indeed, not needed anymore

hamorillo (approved):

PR #556, opened 2026-05-06
RUM-16183: Move from `buildSrc` to isolated build plugins module

0xnm:
### What does this PR do?

This PR moves shared build logic from `buildSrc` to the isolated build plugins module.

Things to note:

* there is a fake `noopBuildConfigClasspath` plugin, it exists just to be able to reference `AndroidConfig` class in samples (applying it brings classes from the plugins module). Kind of hacky solution, but it is just to keep buildscript in samples vanilla (so to avoid bringing `datadogBuildConfig` when there is nothing to configure, and to avoid confusion with `datadog` extension). Another option is to declare these versions in `libs.versions.toml`, which is also a hack imo
* Detekt Public API task fails, because new module should be added to the [exclusion list](https://github.com/DataDog/dd-source/blob/ec50c23850183bde7b481a60c07931ee52339785/domains/mobile/config/android/gitlab/detekt/detekt-public-api.yml)
* Having `Project` in the extension is not quite idiomatic, but for now it is okay. Otherwise the amount of changes would be bigger.
* Most of the buildscript files for samples converted to Kotlin DSL, except one - it stays in Groovy for illustration purpose.


gradle/libs.versions.toml
```diff
@@ -63,15 +62,18 @@ lifecycleRuntimeKtx = "2.8.7"
...
+kotlinGradlePlugin22 = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin22" }
+dokkaGradlePlugin = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version.ref = "dokka" }
+unmockGradlePlugin = { module = "de.mobilej.unmock:UnMockPlugin", version.ref = "unmock" }
+
+versionsGradlePlugin = { module = "com.github.ben-manes:gradle-versions-plugin", version.ref = "versionsPluginGradle" }
 mavenPublishPlugin = { module = "com.vanniktech.maven.publish:com.vanniktech.maven.publish.gradle.plugin", version.ref = "mavenPublishPlugin" }
 
-kotlin = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin" }
+kotlin21 = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin21" }
+kotlin = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin21" }
```
kikoveiga, line 76: Do we need to keep both here?

buildSrc/src/main/kotlin/com/datadog/gradle/Dependencies.kt
```diff
@@ -12,9 +12,4 @@ object Dependencies {
         // Tests Tools
         const val Jacoco = "0.8.7"
     }
-
-    object Repositories {
-        const val Gradle = "https://plugins.gradle.org/m2/"
-        const val Jitpack = "https://jitpack.io"
```
kikoveiga, line 18: Question: why not keep these here instead of hardcoding the strings more than once?

tools/build-config/build.gradle.kts
```diff
@@ -56,21 +56,13 @@ gradlePlugin {
             id = "compilerMetadata" // the alias
             implementationClass = "com.datadog.gradle.plugin.compilermeta.CompilerMetaPlugin"
         }
+        register("datadogBuildConfig") {
```
kikoveiga, line 59: Is it worth to find a way to not hardcode this `datadogBuildConfig` string across different files? Maybe declare it in `libs.versions.toml` `[plugins]` block?

kikoveiga (approved):
Looks good, just some nits. As a follow-up, we will still switch to compiling each KCP module with the respective kotlin version right?

gradle/libs.versions.toml
0xnm: yes, I decided to keep both just to not use alias `kotlin21` in the buildscript of sample apps (and change it with every kotlin version change).

buildSrc/src/main/kotlin/com/datadog/gradle/Dependencies.kt
0xnm: because they are needed in the root `build.gradle.kts` (only Jitpack, Gradle Plugins URL is already a part of the Gradle API) and I cannot reference something in the `build-config` from there without doing tricks, so not worth it.

0xnm:
> As a follow-up, we will still switch to compiling each KCP module with the respective kotlin version right?

I'm not sure if it will be possible, but will see. For now just switching to the included build compared to `buildSrc` for a better isolation.

tools/build-config/build.gradle.kts
0xnm: It is the same approach as for the other plugins above. I'm not sure if `[plugins]` allows declaration without version reference (probably it allows, though), and we have already `buildConfig` there (which is 3rd party config).

ambushwork (approved):

PR #565, opened 2026-06-04
Bump Kotlin version to 2.3.20 for the project

ambushwork:
### What does this PR do?

To be able to compile to incoming kotlin 2.4.0 artifact, we need to Bump Kotlin version from 2.1.21 to 2.3.20.

It is tested that after bumping the version, the app with 2.1.21 can still use the plugin.

### DIFF
--- a/dd-sdk-android-gradle-plugin-kcp-common/api/compiler-meta.txt
+++ b/dd-sdk-android-gradle-plugin-kcp-common/api/compiler-meta.txt
@@ -1,2 +1,2 @@
-kotlin_abi_version=2.1.0
+kotlin_abi_version=2.3.0
 jvm_bytecode_version=11
--- a/dd-sdk-android-gradle-plugin-kcp-common/src/testFixtures/kotlin/com.datadog.gradle.plugin/kcp/KotlinCompilerTest.kt
+++ b/dd-sdk-android-gradle-plugin-kcp-common/src/testFixtures/kotlin/com.datadog.gradle.plugin/kcp/KotlinCompilerTest.kt
@@ -52,6 +52,10 @@ abstract class KotlinCompilerTest {
             // Kotlin 2.4.0 made CommonCompilerArguments.setOptIn non-null; kctfork
             // defaults this property to null, which triggers a NPE in the setter.
             optIn = emptyList()
+            // The project modules are compiled with a newer Kotlin version than the embedded
+            // compiler used in kotlin20/21/22 test modules. Skip the metadata version check
+            // so the embedded compiler can read classes compiled with a newer metadata format.
+            kotlincArguments += "-Xskip-metadata-version-check"
         }.compile()
     }
 
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -7,6 +7,7 @@ build,org.jetbrains.dokka,Apache-2.0,"Copyright 2014-2019 JetBrains s.r.o. and D
 build,org.jetbrains.intellij.deps,LGPL-2.1-only,"Copyright (c) 2001-2002, Eric D. Friedman, Jason Baldridge, Copyright (c) 1999 CERN - European Organization for Nuclear Research"
 import,com.android.tools.build,Apache-2.0,Copyright (C) 2013 The Android Open Source Project
 import,com.squareup.okhttp3,Apache-2.0,"Copyright 2019 Square, Inc"
+import,org.bouncycastle,BouncyCastle-License,"Copyright (c) 2000 - 2026 The Legion of the Bouncy Castle Inc. (https://www.bouncycastle.org)""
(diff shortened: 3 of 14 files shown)


dd-sdk-android-gradle-plugin-kcp-common/src/testFixtures/kotlin/com.datadog.gradle.plugin/kcp/KotlinCompilerTest.kt
```diff
@@ -52,6 +52,10 @@ abstract class KotlinCompilerTest {
             // Kotlin 2.4.0 made CommonCompilerArguments.setOptIn non-null; kctfork
             // defaults this property to null, which triggers a NPE in the setter.
             optIn = emptyList()
+            // The project modules are compiled with a newer Kotlin version than the embedded
+            // compiler used in kotlin20/21/22 test modules. Skip the metadata version check
+            // so the embedded compiler can read classes compiled with a newer metadata format.
+            kotlincArguments += "-Xskip-metadata-version-check"
```
ambushwork, line 58: This line must be added to suppress the error during the compilation test.

In the test of Kotlin21 module for example, Kotlin 2.1 embeddable compiler is used as the test **runtime** compiler, when it reads the meta data of Kotlin 2.3 which is stamped at **build** time, it will throw errors.

Why it doesn't happen on production: `kcp-kotlin21.jar` with kotlin 2.3 meta data is loaded through `-Xplugin=` not `-classpath`, it will not be checked by the client's compiler even though the compiler version is 2.1

dd-sdk-android-gradle-plugin-kcp-common/api/compiler-meta.txt
```diff
@@ -1,2 +1,2 @@
-kotlin_abi_version=2.1.0
+kotlin_abi_version=2.3.0
```
0xnm, line 1: Can you put `languageVersion.set(KotlinVersion.KOTLIN_2_1)` in the `KotlinConfig`? And add a TODO item to bump it later. This way at least we can Metadata API version the same.

dd-sdk-android-gradle-plugin-kcp-common/src/testFixtures/kotlin/com.datadog.gradle.plugin/kcp/KotlinCompilerTest.kt
0xnm: probably with my comment above applied this change is not needed?

dd-sdk-android-gradle-plugin-kcp-common/api/compiler-meta.txt
ambushwork: This doesn't seem to work, because setting `languageVersion.set(KotlinVersion.KOTLIN_2_1)` doesn't change the fact that all the modules are compiled by kotlin 2.3.20 compiler and stamped with 2.3.20 on their metadata.

0xnm: it should

> and stamped with 2.3.20 on their metadata

I did this change with your branch and got `kotlin_abi_version` reset back to 2.1.

Or you mean some other metadata?

ambushwork: yes, I wasn't clear enough, I mean the `@Metadata` annotation in `.class` file, which is actually checked by the compiler of Kotlin Compilation test. but indeed setting `languageVersion` will force the `kotlin_abi_version`

0xnm: `kotlin_abi_version` is generated by parsing `@Metadata` value

ambushwork: OK I was mistaken, after some builds I figure out the whole reason, so the errors of metadata checks come from two parts:
* the `@Metadata` in the .class of our module
* the `@Metadata` in the `kotlin-stdlib.jar`

if we apply `languageVersion = 2.1` on our module, you are right, the `@Metadata` is forced to 2.1,  but `kotlin-stdlib` still has `@Metadata` 2.3, during the testing it still fails.

So there are two workarounds:
* keep `kotlincArguments += "-Xskip-metadata-version-check"` so the compiler in the test ignores all the meta data.
* set `languageVersion = 2.1` and also force the `kotlin-stdlib` to 2.1 in the dependency at the same time.

wdyt?

0xnm: > but `kotlin-stdlib` still has @Metadata 2.3

that is because we have Kotlin as runtime dependency, not as `compileOnly`. I'm wondering if though switching to `compileOnly` is safe.

> and also force the `kotlin-stdlib` to 2.1 in the dependency at the same time

I don't think we should consider this option.

0xnm (approved):

PR #568, opened 2026-06-10
Implement Kotlin 2.4 KCP support for Compose instrumentation

ambushwork:
### What does this PR do?

Implements the `dd-sdk-android-gradle-plugin-kcp-kotlin24` module for Compose instrumentation (NavHost tracking + semantic tag injection) on Kotlin 2.4.x projects. Follows the same per-version sub-plugin pattern as kotlin20/21/22 but required rework due to three breaking API changes in Kotlin 2.4.

Key differences from kotlin20–22

1. `ComponentRegistrar` → `CompilerPluginRegistrar`
Kotlin 2.4 removed `IrGenerationExtension.extensionPointName`, breaking `ComponentRegistrar`-based plugins silently. The kotlin24 `DatadogPluginRegistrarImpl` extends `CompilerPluginRegistrar` directly instead of the shared `DatadogPluginRegistrar` base class.

2. Plugin ordering: LoadingOrder.FIRST → -Xcompiler-plugin-order
CompilerPluginRegistrar doesn't support LoadingOrder. Without ordering, Compose runs first and Datadog produces invalid IR. DatadogKotlinCompilerPluginSupport now injects -Xcompiler-plugin-order=com.datadoghq.kotlin.compiler>androidx.compose.compiler.plugins.kotlin for Kotlin 2.4 compilations.

3. `NavHost` tracking: `apply` lambda → inline `irBlock` with all args explicit
Compose 2.4 no longer generates `$default` static overloads for `@Composable` functions — null IR argument slots that previously triggered default-filling now cause `NoSuchMethodError` at runtime. The `kotlin22` approach injects `navController.apply { InstrumentedNavigationViewTrackingEffect(this) }` with one arg. The kotlin24 transformer emits an inline `irBlock` with all four arguments set explicitly, and guards `initReferences()` to abort with a clear warning if any required symbol (`AcceptAllNavDestinations`, `Datadog.getInstance()`) is unresolvable.

4. Argument indexing: `putValueArgument(valueIndex)` → `arguments[unifiedIndex]`
Kotlin 2.4 unifies dispatch receiver, extension receiver, and regular parameters into a single arguments[] array. Both transformers use `indexOfFirst { it.kind == ... }` on the full parameters list instead of the deprecated `putValueArgument(i)` / `getValueArgument(i)` APIs.

Also changed

- `InstrumentationMode.from()` and `DatadogPluginRegistrar.CONFIG_INSTRUMENTATION_MODE` made public so the `kotlin24` registrar (which doesn't extend the base class) can access them.
- `KotlinCompilerTestSources` updated with the real 4-parameter signature of `InstrumentedNavigationViewTrackingEffect` and stub source files for `SdkCore`, `AcceptAllNavDestinations`, and `Datadog`.

### DIFF
--- a/dd-sdk-android-gradle-plugin-kcp-common/src/testFixtures/kotlin/com.datadog.gradle.plugin/kcp/KotlinCompilerTestSources.kt
+++ b/dd-sdk-android-gradle-plugin-kcp-common/src/testFixtures/kotlin/com.datadog.gradle.plugin/kcp/KotlinCompilerTestSources.kt
@@ -18,6 +18,10 @@ object KotlinCompilerTestSources {
     const val TRACKING_EFFECT_FILE_NAME = "Navigation.kt"
     const val DD_MODIFIER_CLASS_FILE_NAME = "DatadogModifier.kt"
     const val COMPOSE_INSTRUMENTATION_ANNOTATION_FILE_NAME = "ComposeInstrumentation.kt"
+    const val SDK_CORE_FILE_NAME = "SdkCore.kt"
+    const val COMPONENT_PREDICATE_FILE_NAME = "ComponentPredicate.kt"
+    const val ACCEPT_ALL_NAV_DESTINATIONS_FILE_NAME = "AcceptAllNavDestinations.kt"
+    const val DATADOG_FILE_NAME = "Datadog.kt"
 
     @Language("kotlin")
     val DD_MODIFIER_SOURCE_FILE_CONTENT =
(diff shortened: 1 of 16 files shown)


ambushwork:
@codex review

dd-sdk-android-gradle-plugin-kcp-common/src/testFixtures/kotlin/com.datadog.gradle.plugin/kcp/KotlinCompilerTestSources.kt
```diff
@@ -55,17 +59,60 @@ object KotlinCompilerTestSources {
...
         import com.datadog.gradle.plugin.kcp.TestCallbackContainer
 
-
         @Composable
         fun InstrumentedNavigationViewTrackingEffect(
-            navController: NavController
+            navController: NavController,
+            trackArguments: Boolean = true,
+            destinationPredicate: ComponentPredicate<NavDestination> = AcceptAllNavDestinations(),
+            sdkCore: SdkCore = com.datadog.android.Datadog.getInstance()
```
0xnm: prefer `import`

dd-sdk-android-gradle-plugin-kcp-kotlin24/src/main/kotlin/com/datadog/gradle/plugin/kcp/DatadogPluginRegistrarImpl.kt
```diff
@@ -4,9 +4,65 @@
  * Copyright 2020-Present Datadog, Inc.
  */
 
+@file:OptIn(FirIncompatiblePluginAPI::class)
```
0xnm: why it is needed? why it is incompatible?

0xnm (dismissed):

dd-sdk-android-gradle-plugin-kcp-common/src/testFixtures/kotlin/com.datadog.gradle.plugin/kcp/KotlinCompilerTestSources.kt
ambushwork: updated.

0xnm (approved):

dd-sdk-android-gradle-plugin-kcp-kotlin24/src/main/kotlin/com/datadog/gradle/plugin/kcp/DatadogPluginRegistrarImpl.kt
ambushwork: this is copy paste from previous version, can be removed here

Issue #575, opened 2026-06-12
Compose instrumentation silently corrupts first-frame composition on Kotlin 2.3 (kotlin22 KCP applied to 2.3 compiler)


achatikyan:
### Describe what happened

With `composeInstrumentation = InstrumentationMode.AUTO` on Kotlin 2.3.20, instrumented composables render incorrectly on the **first composition frame**: `TextField` values render empty, fields collapse to intrinsic width, and content shifts. Real devices recover on the next recomposition, so the app looks fine at runtime, but any single frame capture fails: in our app 62 Paparazzi screenshot goldens broke with pixel diffs between 0.6% and 3.2%, showing text fields losing their passed values (expected golden shows the populated value, actual render shows the field empty and collapsed).

This appears to be the successor of #474 and #520: those crashes were resolved by making the kotlin22 compiler plugin load on Kotlin 2.3, but the result is a 2.2-built KCP running inside the 2.3 compiler, which now compiles successfully while producing subtly malformed IR.

### Versions

```
kotlin: 2.3.20
datadog-gradle-plugin: 1.26.0 (also reproduced on 1.27.0)
dd-sdk-android: 3.10.0
composeBom: 2026.03.01
agp: 9.2.1
paparazzi: 2.0.0-alpha04
```

### Steps to reproduce

1. Android library module with Compose UI and Paparazzi screenshot tests, Kotlin 2.3.20
2. Apply the plugin with `composeInstrumentation = InstrumentationMode.AUTO`
3. Run Paparazzi verification against goldens recorded without instrumentation
4. Tests fail; composables containing `TextField`s with passed values render them empty on the captured frame

`InstrumentationMode.ANNOTATION` reproduces it as well, since the KCP is attached for any mode other than `DISABLE`. `DISABLE` is clean.

### Evidence isolating the cause

* AUTO fails, DISABLE passes, on clean builds both locally and in CI with identical diffs
* `javap` on the affected file shows the only change under AUTO is the injected `instrumentedDatadog` + `Modifier.then` calls (37 in one file)
* 1.27.0 (built against Kotlin 2.3.20 per the changelog) reproduces identically, so this is not just a build-toolchain skew
* Patching `KotlinVersion.from()` so 2.3 returns `UNSUPPORTED` (plugin no-ops instead of attaching the KCP), then requesting AUTO: compilation contains zero injected calls and all 62 goldens pass

### Root cause

[...]

PR #577, opened 2026-06-15
RUM-16787: Preserve callee Modifier default when argument is omitted

ambushwork:
### What does this PR do?

Fixes a Compose instrumentation bug (#575) where the `ComposeTagTransformer` could **silently corrupt layouts** by discarding a composable's default `Modifier`.

This PR makes the transformer preserve the callee's default in those cases: it only overwrites an omitted `Modifier` argument when the default is a bare companion (or a deserialized placeholder from a dependency). Otherwise it leaves the call site untouched so the real default applies. The change is applied across the
`kotlin20/21/22/24` KCP transformers.

### In which use cases it didn't work (now fixed)

**1. Nullable `Modifier?` default consumed via an elvis fallback** — the real-world trigger from the report:

```kotlin
@Composable
fun IconRow(text: String, iconModifier: Modifier? = null) {
    Box(modifier = (iconModifier ?: Modifier.size(40.dp)).background(Color.Red))
}

// caller omits iconModifier:
IconRow(text = "…")
```

- **Before:** the transformer injected `Modifier.datadog("IconRow")` into the omitted
  `iconModifier`, making it **non-null**. The `?:` elvis short-circuited to the left,
  so `Modifier.size(40.dp)` was never applied — the `Box` collapsed to 0×0 and the
  background disappeared.
- **After:** `iconModifier` stays `null`, the elvis falls back to `Modifier.size(40.dp)`,
  and the layout renders correctly.

**2. Non-null, behavior-carrying default** (e.g. a sizing modifier):

```kotlin
@Composable
fun IconRow(text: String, iconModifier: Modifier = Modifier.size(40.dp)) {
    Box(modifier = iconModifier.background(Color.Red))
}

IconRow(text = "…") // iconModifier omitted
```

- **Before:** the `Modifier.size(40.dp)` default was overwritten with
  `Modifier.datadog(...)`, dropping the size.
- **After:** the `Modifier.size(40.dp)` default is preserved.

Both reproduced on Kotlin 2.2, 2.3, and 2.4 (kotlin22 and kotlin24 KCP modules), so this is a transform-level issue, not a compiler-version skew.

### Trade-off

A user composable that opts into a non-bare modifier default (e.g. `Modifier? = null`) and is called without a modifier no longer receives an auto-tag on **that** parameter slot. Its contents are still tagged (nested calls with present/companion-default modifiers are unaffected). This is the necessary trade to keep the layout correct.

### Motivation

RUM-16787 / #575

### Tests

- **KCP compilation tests** (run per version, kotlin20/21/22/24) — compile and execute the
  transformed bytecode:
  - omitted nullable default consumed via an elvis → fallback preserved;
  - omitted non-null sizing default → default preserved;
  - omitted modifier on a **dependency** composable (deserialized default) → still instrumented (regression guard);
  - existing cases (bare companion / explicit / custom modifier) unchanged.
- **Instrumented (on-device) tests** in `:instrumented` — real Compose + plugin in AUTO mode,
  asserting via Compose semantics that the preserved default's `testTag` survives while the
  inner `Box` still carries the `_dd_semantics` tag.

### DIFF
--- a/dd-sdk-android-gradle-plugin-kcp-kotlin20/src/main/kotlin/com/datadog/gradle/plugin/kcp/ComposeTagTransformer.kt
+++ b/dd-sdk-android-gradle-plugin-kcp-kotlin20/src/main/kotlin/com/datadog/gradle/plugin/kcp/ComposeTagTransformer.kt
@@ -18,8 +18,10 @@ import org.jetbrains.kotlin.ir.builders.irString
 import org.jetbrains.kotlin.ir.declarations.IrDeclaration
 import org.jetbrains.kotlin.ir.declarations.IrFunction
 import org.jetbrains.kotlin.ir.declarations.IrPackageFragment
+import org.jetbrains.kotlin.ir.declarations.IrValueParameter
 import org.jetbrains.kotlin.ir.expressions.IrCall
 import org.jetbrains.kotlin.ir.expressions.IrComposite
+import org.jetbrains.kotlin.ir.expressions.IrConst
 import org.jetbrains.kotlin.ir.expressions.IrExpression
 import org.jetbrains.kotlin.ir.expressions.IrFunctionExpression
 import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
@@ -121,8 +123,15 @@ constructor(
         expression.symbol.owner.valueParameters.forEachIndexed { index, irValueParameter ->
             // Locate where Modifier is accepted in the parameter list and replace it with the new expression.
             if (irValueParameter.type.classFqName == modifierClassFqName) {
+                val argument = expression.getValueArgument(index)
+                // Skip instrumenting an omitted argument whose callee default carries behavior
+                // (e.g. `Modifier.size(...)` or `null` consumed via an elvis). Overwriting it would
+                // discard that default and corrupt the layout. See #575.
+                if (shouldPreserveCalleeDefault(irValueParameter, argument)) {
+                    return@forEachIndexed
+                }
                 val irExpression = buildIrExpression(
-                    expression = expression.getValueArgument(index),
+                    expression = argument,
                     builder = builder,
                     functionName = expression.symbol.owner.name.asString(),
                     isImageComposableFunction = isImageComposableFunction(expression)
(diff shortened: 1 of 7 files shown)


ambushwork:
@codex review

dd-sdk-android-gradle-plugin-kcp-kotlin20/src/main/kotlin/com/datadog/gradle/plugin/kcp/ComposeTagTransformer.kt
```diff
@@ -217,6 +226,22 @@ constructor(
...
+        irValueParameter: IrValueParameter,
+        argument: IrExpression?
+    ): Boolean {
+        val isArgumentAbsent = argument == null ||
+            (argument is IrComposite && argument.type.classFqName == kotlinNothingFqName)
+        if (!isArgumentAbsent) {
+            return false
+        }
+        val defaultExpression = irValueParameter.defaultValue?.expression
+        return defaultExpression != null && !defaultExpression.isModifierCompanionReference()
```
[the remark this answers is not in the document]
ambushwork: valid point, fixed

aleksandr-gringauz (approved):

dd-sdk-android-gradle-plugin-kcp-kotlin20/src/main/kotlin/com/datadog/gradle/plugin/kcp/ComposeTagTransformer.kt
```diff
@@ -217,6 +226,28 @@ constructor(
...
+        if (!isArgumentAbsent) {
+            return false
+        }
+        // Only preserve the callee default when we can positively identify it as a value that carries
+        // behavior. A bare `Modifier` companion (`= Modifier`) is an IrGetObjectValue and is safe to
+        // overwrite; defaults deserialized from dependencies arrive as IrErrorExpression placeholders and
+        // must also be instrumented. Both fall through to `false` here. See #575.
+        return when (irValueParameter.defaultValue?.expression) {
+            // e.g. `Modifier? = null`, consumed via an `iconModifier ?: <fallback>` elvis in the body.
+            is IrConst<*> -> true
```
kikoveiga, line 244: I see in the other modules `is IrConst -> true` is used without the `<*>`, is it intended?

kikoveiga (approved):

dd-sdk-android-gradle-plugin-kcp-kotlin20/src/main/kotlin/com/datadog/gradle/plugin/kcp/ComposeTagTransformer.kt
ambushwork: Good question, yes, because each modules depends on the different versions of `kotlin-compiler-embeddable`, in version 2.0.21 `IrConst` is defined with wildcard `abstract class IrConst<T>`, and `abstract class IrConst` without wildcard since version 2.1.0.

PR #602, opened 2026-07-28
RUM-17794: Drop support of AGP versions below 8

0xnm:
### What does this PR do?

[AGP 8.0](https://developer.android.com/build/releases/agp-8-0-0-release-notes) was released more than 3 years ago, so we can safely drop support of the versions below.

AGP 8.0 requires also having JVM bytecode version 17, so bumping it as well.

### DIFF
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/CurrentAgpVersion.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/CurrentAgpVersion.kt
@@ -11,18 +11,12 @@ import com.datadog.gradle.plugin.TaskUtils
 @Suppress("MagicNumber")
 internal object CurrentAgpVersion {
 
+    val CAN_USE_PLUGIN: Boolean
+        get() = TaskUtils.isAgpEqualOrAbove(major = 8, minor = 0, patch = 0)
+
     // can work probably even with lower versions, but legacy Variant API is working fine there as well
     val CAN_ENABLE_NEW_VARIANT_API: Boolean
         get() = TaskUtils.isAgpEqualOrAbove(major = 8, minor = 4, patch = 0)
-
-    val SUPPORTS_KOTLIN_DIRECTORIES_SOURCE_PROVIDER: Boolean
-        get() = TaskUtils.isAgpEqualOrAbove(major = 7, minor = 0, patch = 0)
-
-    val EXTERNAL_NATIVE_BUILD_SOFOLDER_IS_PUBLIC: Boolean
-        get() = TaskUtils.isAgpEqualOrAbove(major = 8, minor = 0, patch = 0)
-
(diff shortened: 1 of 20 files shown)


dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/CurrentAgpVersion.kt
```diff
@@ -11,19 +11,16 @@ import com.datadog.gradle.plugin.TaskUtils
...
+        get() = TaskUtils.isAgpEqualOrAbove(major = 8, minor = 0, patch = 0)
+
     // can work probably even with lower versions, but legacy Variant API is working fine there as well
     val CAN_ENABLE_NEW_VARIANT_API: Boolean
         get() = TaskUtils.isAgpEqualOrAbove(major = 8, minor = 4, patch = 0)
 
-    val SUPPORTS_KOTLIN_DIRECTORIES_SOURCE_PROVIDER: Boolean
-        get() = TaskUtils.isAgpEqualOrAbove(major = 7, minor = 0, patch = 0)
-
     val EXTERNAL_NATIVE_BUILD_SOFOLDER_IS_PUBLIC: Boolean
```
hamorillo: nit: I think we can remove this one (and the associated if) as it will be always true.

hamorillo (dismissed):

dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/DdAndroidGradlePluginFunctionalTest.kt
```diff
@@ -1706,12 +1706,12 @@ internal class DdAndroidGradlePluginFunctionalTest {
...
             BuildVersionConfig(
-                agpVersion = "7.0.4",
-                gradleVersion = "7.4",
+                agpVersion = "8.0.0",
+                gradleVersion = "8.0",
                 buildToolsVersion = "31.0.0",
                 targetSdkVersion = "31",
-                kotlinVersion = "1.6.10",
-                jvmTarget = JavaVersion.VERSION_11.toString()
+                kotlinVersion = "1.7.20",
```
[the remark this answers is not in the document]
0xnm: That is false, Kotlin 1.7 works fine with Gradle 8.

dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/internal/CurrentAgpVersion.kt
0xnm: Good point! Updated.

hamorillo (approved):

ambushwork (approved):

kikoveiga (commented):
Should we add some test to make sure the plugin won't be applied with `AGP < 8`?

PR #605, opened 2026-07-28
[RUM-16312] Support gzip-compressed JVM mapping file uploads

buranmert:
### What does this PR do?

Adds an opt-in `dd-compress-mapping-file` Gradle property. When set, the JVM/Android mapping file content is gzip-compressed before being attached to the upload request, and a `compression: "gzip"` field is added to the `event` metadata JSON so the intake/`sourcemap-processor` knows to decompress it.

Only the JVM mapping file upload is affected (NDK symbol file uploads are untouched), matching backend support which is JVM-mapping-specific.

### Motivation

Backend support for accepting gzip-compressed JVM/Android mapping files was added in `dd-go` PR [#2682](https://github.com/ddoghq/dd-go/pull/2682) (`sourcemap-processor` now decompresses the mapping blob in memory when `Metadata.Compression == "gzip"`). ProGuard/R8 mapping files are highly compressible text (often 10-20x), so compressing the upload reduces both upload payload size and stored blob size, and helps users hit the existing max-mapping-file-size limit less often.

The feature is kept **opt-in** (not default-on) since the backend rollout may lag across Datadog sites/regions; a site whose `sourcemap-processor` doesn't yet understand `compression: gzip` would otherwise store the mapping file permanently gzip-compressed, breaking deobfuscation. The default can be flipped in a later release once the backend is confirmed rolled out everywhere.

### Additional Notes

- `Uploader.UploadFileInfo` gained a `compressed: Boolean` field, driven per-call-site rather than hardcoded by file type, so only `MappingFileUploadTask` sets it (gated by the new Gradle property).
- `OkHttpUploader` gzip-compresses the file part content (independent of the existing `useGzip`/`Content-Encoding: gzip` transport-level compression, which is unrelated) and adds the `compression` event field. Compression is **streamed** (via the existing `RequestBody.gzip()` wrapper, reused for a single multipart part) rather than loading the whole mapping file into memory, so peak memory stays flat regardless of file size — see benchmarks below.
- Tests: `OkHttpUploaderTest` verifies the uploaded file part is actually gzip-compressed and decompresses back to the original content (including a combined case with the outer transport-level gzip also enabled, gunzipping both layers and diffing against the original file), and that the `compression` field is present in the event JSON. `DdAndroidGradlePluginFunctionalTest` verifies the Gradle property wiring end-to-end on a real `assembleRelease` build.
- Documented the new opt-in property in the README, under "Compressing the mapping file upload".

### Benchmarks

Benchmarked the real `OkHttpUploader#upload` code path end-to-end: a real OkHttp request/response round-trip through a local `MockWebServer` bound to `127.0.0.1` (no real/remote endpoint is ever contacted). Run as a standalone JVM process outside Gradle to avoid Gradle/JVM-startup noise. Platform: macOS/arm64, JDK 17, using `/usr/bin/time -l` (peak RSS) and `/usr/bin/time -p` (wall time).

**Wall time**

| Mapping file size | No compression | With compression | Δ |
|---|---|---|---|
| 20 MB | ~0.68s | ~0.76s | +0.08s |
| 150 MB | ~0.78s | ~1.53s | +0.75s |

Compression adds real, roughly file-size-linear CPU time. In the real upload flow this is traded against a much smaller payload to transmit (at ~23x compression, a 150 MB upload becomes ~6.5 MB), so on any bandwidth-constrained connection the transfer-time savings should outweigh the added compression time.

**Peak RSS**

| Mapping file size | No compression | With compression | Δ |
|---|---|---|---|
| 20 MB | ~228 MB | ~229 MB | noise (no clear signal) |
| 150 MB | ~370 MB | ~225 MB | **-145 MB (compression uses *less* memory)** |

At 20 MB the ~220 MB of fixed JVM/OkHttp/Mockito/MockWebServer class-loading overhead dominates and swamps any signal. At 150 MB, compression comes out clearly *ahead* on memory too: `MockWebServer` fully buffers whatever bytes actually cross the wire as part of recording the request, so the uncompressed run has to hold the full ~150 MB mapping file in memory on top of the baseline, while the compressed run only holds the ~6.5 MB gzip output. In other words, once you account for what has to be buffered on the receiving side (a proxy, the real intake service, etc.), not just the sending side, compression is a net memory win, not just a network-payload win. An isolated micro-benchmark of only the client-side gzip step (excluding anything downstream) showed a small, flat, streaming-related overhead instead (+1-1.5 MB regardless of file size) — that number is real too, but it's a narrower measurement than this end-to-end one.

<details>
<summary>Reproduction steps</summary>

`BenchMain.kt` below drives the actual `OkHttpUploader` class directly (not a reimplementation). It's not committed to the repo (it's a throwaway benchmarking entry point, not a test), so drop it into the test source set locally to reproduce:

1. Save it as `dd-sdk-android-gradle-plugin/src/test/kotlin/com/datadog/gradle/plugin/internal/BenchMain.kt`.
2. Compile it as part of the module (so it can see `internal` declarations) and grab the test runtime classpath:
   ```bash
   ./gradlew --configure-on-demand :dd-sdk-android-gradle-plugin:compileTestKotlin

   cat > /tmp/print_classpath.gradle.kts <<'EOF'
   allprojects {
       if (project.path == ":dd-sdk-android-gradle-plugin") {
           afterEvaluate {
               tasks.register("printTestClasspath") {
                   doLast {
                       val testSourceSet = project.extensions.getByType(org.gradle.api.tasks.SourceSetContainer::class.java).getByName("test")
                       val cp = testSourceSet.runtimeClasspath.files.joinToString(":") { it.absolutePath }
                       java.io.File("/tmp/bench/testcp.txt").writeText(cp)
                   }
               }
           }
       }
   }
   EOF
   mkdir -p /tmp/bench
   ./gradlew --configure-on-demand -I /tmp/print_classpath.gradle.kts :dd-sdk-android-gradle-plugin:printTestClasspath
   ```
3. Generate a large, highly-compressible synthetic mapping file:
   ```bash
   python3 - <<'PYEOF'
   lines = []
   size = 0
   i = 0
   while size < 150 * 1024 * 1024:
       header = f"com.example.foo.Class{i % 5000} -> a.a.{i}:\n"
       lines.append(header)
       size += len(header)
       for j in range(20):
           member = f"    {j*3+1}:{j*3+3}:void method{j}(int,java.lang.String) -> {chr(97+j%26)}\n"
           lines.append(member)
           size += len(member)
       i += 1
   open("/tmp/bench/mapping.txt", "w").writelines(lines)
   PYEOF
   ```
4. Run it directly as a plain `java` process (bypassing Gradle's test runner so `/usr/bin/time` can wrap the exact process):
   ```bash
   CP="dd-sdk-android-gradle-plugin/build/classes/kotlin/main:dd-sdk-android-gradle-plugin/build/classes/kotlin/test:$(cat /tmp/bench/testcp.txt)"

   # RSS (peak resident set size)
   /usr/bin/time -l java -cp "$CP" com.datadog.gradle.plugin.internal.BenchMainKt streamed /tmp/bench/mapping.txt   # no compression
   /usr/bin/time -l java -cp "$CP" com.datadog.gradle.plugin.internal.BenchMainKt compressed /tmp/bench/mapping.txt # with compression

   # Wall time
   /usr/bin/time -p java -cp "$CP" com.datadog.gradle.plugin.internal.BenchMainKt streamed /tmp/bench/mapping.txt
   /usr/bin/time -p java -cp "$CP" com.datadog.gradle.plugin.internal.BenchMainKt compressed /tmp/bench/mapping.txt
   ```

`BenchMain.kt`:
```kotlin
package com.datadog.gradle.plugin.internal

import com.datadog.gradle.plugin.DatadogSite
import okhttp3.mockwebserver.MockResponse
import okhttp3.mockwebserver.MockWebServer
import org.mockito.kotlin.doReturn
import org.mockito.kotlin.mock
import org.mockito.kotlin.whenever
import java.io.File

// Standalone entry point (not a @Test) to benchmark the real OkHttpUploader#upload code path
// (RSS / wall time) outside of Gradle's own overhead. Run with `java -cp <classpath> BenchMainKt
// <streamed|compressed> <path-to-mapping-file>`, wrapped in `/usr/bin/time`.
//
// Uploads only ever go to a local MockWebServer bound to 127.0.0.1 -- no real network call is
// made and no real Datadog endpoint is ever contacted.
fun main(args: Array<String>) {
    val mode = args[0]
    val path = args[1]

    val mockWebServer = MockWebServer()
    mockWebServer.enqueue(MockResponse().setResponseCode(200).setBody("{}"))
    mockWebServer.start()

    val mockSite: DatadogSite = mock()
    whenever(mockSite.uploadEndpoint()) doReturn mockWebServer.url("/upload").toString()

    val fileInfo = Uploader.UploadFileInfo(
        fileKey = "jvm_mapping_file",
        file = File(path),
        encoding = "text/plain",
        fileType = "jvm_mapping_file",
        fileName = "jvm_mapping",
        compressed = mode == "compressed"
    )

    val identifier = DdAppIdentifier(
        serviceName = "bench-service",
        version = "1.0",
        versionCode = 1,
        variant = "release",
        buildId = "bench-build-id"
    )

    OkHttpUploader().upload(
        site = mockSite,
        fileInfo = fileInfo,
        repositoryFile = null,
        apiKey = "fake-api-key",
        identifier = identifier,
        repositoryInfo = null,
        useGzip = false,
        emulateNetworkCall = false
    )

    mockWebServer.shutdown()
    println("done mode=$mode")
}
```

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,24 @@ Similarly, to upload NDK symbols, run the `uploadNdkSymbolFiles[Variant]` task i
 ./gradlew uploadNdkSymbolFilesRelease
 ```
 
+#### Compressing the mapping file upload
+
+If your mapping file is large enough to hit the upload size limit, you can opt in to gzip-compressing
+it before upload by setting the `dd-compress-mapping-file` Gradle property, either in `gradle.properties`:
+
+```properties
+dd-compress-mapping-file=true
+```
+
+or on the command line when running the upload task:
+
+```bash
+./gradlew uploadMappingRelease -Pdd-compress-mapping-file
+```
+
+This is opt-in (not enabled by default) since it requires your Datadog site to support decompressing
+gzip-compressed mapping files.
+
 ### Configuration
 
 You can configure the plugin by adding the following block at the end of your `build.gradle` file.
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ or on the command line when running the upload task:
 This is opt-in (not enabled by default) since it requires your Datadog site to support decompressing
 gzip-compressed mapping files.
 
+This property is independent of the transport-level compression that uploads already use by default
+(`Content-Encoding: gzip`, turned off with `dd-disable-gzip`). Uploads are therefore already
+compressed in transit without it; what this property changes is that the mapping file itself arrives
+compressed, rather than only the connection carrying it being compressed.
+
 ### Configuration
 
 You can configure the plugin by adding the following block at the end of your `build.gradle` file.
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/FileUploadTask.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/FileUploadTask.kt
@@ -59,6 +59,19 @@ abstract class FileUploadTask @Inject constructor(
     private val disableGzipOption: Provider<String> =
         providerFactory.gradleProperty(DISABLE_GZIP_GRADLE_PROPERTY)
 
+    // TODO RUM-16312 Remove this opt-in gate and enable by default once all Datadog sites support
+    //  decompressing gzip-compressed JVM mapping files.
+    private val compressMappingFileOption: Provider<String> =
+        providerFactory.gradleProperty(COMPRESS_MAPPING_FILE_GRADLE_PROPERTY)
+
+    /**
+     * Whether the JVM/Android mapping file content should be gzip-compressed before upload.
+     * Opt-in via the [COMPRESS_MAPPING_FILE_GRADLE_PROPERTY] Gradle property.
+     */
+    @get:Internal
+    protected val compressMappingFile: Boolean
+        get() = compressMappingFileOption.isPresent
+
     // needed for functional tests, because we don't have real API key
     private val emulateNetworkCall: Provider<String> =
         providerFactory.gradleProperty(EMULATE_UPLOAD_NETWORK_CALL)
@@ -287,6 +300,7 @@ abstract class FileUploadTask @Inject constructor(
         internal val LOGGER = Logging.getLogger("DdFileUploadTask")
 
         const val DISABLE_GZIP_GRADLE_PROPERTY = "dd-disable-gzip"
+        const val COMPRESS_MAPPING_FILE_GRADLE_PROPERTY = "dd-compress-mapping-file"
         const val EMULATE_UPLOAD_NETWORK_CALL = "dd-emulate-upload-call"
 
         const val API_KEY_MISSING_ERROR = "Make sure you define an API KEY to upload your mapping files to Datadog. " +
--- a/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/FileUploadTask.kt
+++ b/dd-sdk-android-gradle-plugin/src/main/kotlin/com/datadog/gradle/plugin/FileUploadTask.kt
@@ -59,17 +59,6 @@ abstract class FileUploadTask @Inject constructor(
     private val disableGzipOption: Provider<String> =
         providerFactory.gradleProperty(DISABLE_GZIP_GRADLE_PROPERTY)
 
-    private val compressMappingFileOption: Provider<String> =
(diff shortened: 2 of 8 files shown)


buranmert:
Superseded by #607, opened directly against upstream now that I have push access to this repo.

PR #619, opened 2026-09-01
Improve CI pipeline and image maintenance

kikoveiga:
### Summary

- Replaces the shared static-analysis pipeline with repository-owned Detekt, Ktlint, Android Lint, license, and generated-file checks aligned with `dd-sdk-android`.
- Centralizes Ktlint configuration, standardizes source headers, and converts the remaining Groovy build script to Kotlin DSL.
- Replaces JaCoCo and its flaky per-module coverage threshold with Kover 0.9.8 while preserving module-level Datadog coverage uploads.
- Streamlines the CI image and build context, embeds the checksum-verified Java tracer, updates Android tooling, and supports architecture-aware Datadog CI installation.
- Aligns pipeline stage naming and auto-cancels superseded short-lived branch pipelines while protecting stateful jobs.

### Validation

- `./gradlew koverXmlReport :build-config:test :dd-sdk-android-gradle-plugin:checkThirdPartyLicences`
- `./gradlew ktlintCheck :build-config:ktlintCheck detekt --continue`
- `./gradlew checkGeneratedFiles --no-configuration-cache`
- `./gradlew lintRelease --continue`
- `./gradlew :samples:variants:tasks`
- GitLab YAML and diff formatting checks

### Notes

- These are CI and build-tooling changes; no plugin version bump is required.
- The manual `ci-image` job must publish and sign image `19`.
- The full image build was not run locally because Docker was unavailable.


ci/Dockerfile.gitlab
```diff
@@ -1,56 +1,35 @@
+# syntax=docker/dockerfile:1
```
0xnm: what is this line for? we don't have such in `dd-sdk-android` repo

ci/Dockerfile.gitlab
```diff
@@ -60,39 +39,31 @@ RUN mkdir /root/.android \
...
+    echo y | /android-sdk-linux/cmdline-tools/latest/bin/sdkmanager --install "cmake;${CMAKE_VERSION}" >/dev/null && \
+    (yes || true) | /android-sdk-linux/cmdline-tools/latest/bin/sdkmanager --licenses
+
+ENV ANDROID_SDK_ROOT=/android-sdk-linux \
+    ANDROID_HOME=/android-sdk-linux \
+    ANDROID_NDK=/android-sdk-linux/ndk/${NDK_VERSION}
+ENV PATH=${PATH}:${ANDROID_HOME}/platform-tools:${ANDROID_SDK_ROOT}/build-tools/${ANDROID_BUILD_TOOLS}:${ANDROID_NDK}
+
+# Install Datadog Java tracer
+ENV DD_TRACER_VERSION=1.54.0 \
```
0xnm: maybe use the latest version which is 1.65.1? if it doesn't break anything

ci/pipelines/default-pipeline.yml
```diff
@@ -3,11 +3,23 @@ include:
...
+      auto_cancel:
+        on_new_commit: none
+
 stages:
   - ci-image
   - security
-  - analysis
-  - test
+  - verify
+  - test-pyramid
```
0xnm, line 22: we don't really have "test pyramid" notion here, because we have only 2 layers: unit tests + instrumented

but we can keep it

ci/pipelines/default-pipeline.yml
```diff
@@ -74,38 +88,45 @@ create_key:
...
   image: $CI_IMAGE_DOCKER
-  stage: analysis
+  stage: verify
   timeout: 30m
+
+analysis:generated-files-check:
+  extends: .verification-job
   script:
-    - GRADLE_OPTS="-Xmx2560m" ./gradlew :dd-sdk-android-gradle-plugin:checkThirdPartyLicences --stacktrace --no-daemon
+    - ./gradlew checkGeneratedFiles --stacktrace --no-daemon --no-configuration-cache
```
0xnm: what breaks configuration cache here?

ci/pipelines/default-pipeline.yml
```diff
@@ -74,38 +88,45 @@ create_key:
...
+    - ./gradlew lintRelease --continue --stacktrace --no-daemon
+
+analysis:licenses:
+  extends: .verification-job
+  script:
+    - ./gradlew :dd-sdk-android-gradle-plugin:checkThirdPartyLicences --stacktrace --no-daemon
 
 # TESTS
 
 test:plugin:
```
0xnm, line 126: Just noticed that we have the same in Android SDK repo: shouldn't we match task prefix with a stage name? but probably it doesn't matter

samples/variants/build.gradle
0xnm, line 1: This was in Groovy on purpose, I guess we need either revert this change or admit that we don't acknowledge plugin usage from Groovy build script and have a single `samples/variants` flavor which is backed by Kotlin buildscript only (which is `samples/variants-kotlin` currently).

ci/Dockerfile.gitlab
kikoveiga: True, it's not needed. Removed it.

ci/Dockerfile.gitlab
kikoveiga: Good point, updated.

samples/variants/build.gradle
kikoveiga: Sorry about that, we should revert then.

ci/pipelines/default-pipeline.yml
kikoveiga: Yes, this is just to align with `dd-sdk-android`.

ci/pipelines/default-pipeline.yml
kikoveiga: It's because we were accessing into `project`. Did a small refactor in `TransitiveDependenciesPlugin` & `TransitiveDependenciesTask` and managed to remove the `--no-configuration-cache` flag. Still, I think configuration cache is in general disabled for the project, could try that in an upcoming PR.

ci/pipelines/default-pipeline.yml
kikoveiga: Yeah makes sense, but this way we also distinguish between `test` and `analysis` jobs. I think @abrooksv joined these to improve the speed right?

ci/pipelines/default-pipeline.yml
```diff
@@ -74,38 +88,45 @@ create_key:
...
 
-analysis:woke:
-  tags: [ "arch:amd64" ]
-  image: $CI_IMAGE_DOCKER
-  stage: analysis
-  timeout: 30m
+analysis:ktlint:
+  extends: .verification-job
+  script:
+    - ./gradlew ktlintCheck :build-config:ktlintCheck --stacktrace --no-daemon
```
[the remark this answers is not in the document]
kikoveiga: Fixed.

ci/pipelines/default-pipeline.yml
```diff
@@ -74,38 +88,45 @@ create_key:
...
+analysis:detekt:
+  extends: .verification-job
+  script:
+    - ./gradlew detekt --continue --stacktrace --no-daemon
+
+analysis:android-lint:
+  extends: .verification-job
   script:
-    - /usr/bin/woke --exit-1-on-failure
+    - ./gradlew lintRelease --continue --stacktrace --no-daemon
```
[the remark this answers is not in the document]
kikoveiga: Fixed, changed `lintRelease` to `lint`.

ci/pipelines/default-pipeline.yml
abrooksv: Hey, I joined them for parallelization since the jobs were all less than the long pole of the stage so it actually shortened the total pipeline duration.

The names were kept due to they are used in the CI Viz product so it kept the metrics for data over time from becoming split/reset.

kikoveiga: Thanks for the insights!

tools/build-config/src/main/kotlin/com/datadog/gradle/plugin/transdeps/TransitiveDependenciesPlugin.kt
```diff
@@ -9,16 +9,29 @@ package com.datadog.gradle.plugin.transdeps
...
 import org.gradle.api.Project
+import org.gradle.api.artifacts.component.ProjectComponentIdentifier
 import org.gradle.kotlin.dsl.register
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 class TransitiveDependenciesPlugin : Plugin<Project> {
 
     override fun apply(target: Project) {
+        val runtimeClasspathConfiguration = target.configurations.getByName("runtimeClasspath")
+        check(runtimeClasspathConfiguration.isCanBeResolved) { "$runtimeClasspathConfiguration cannot be resolved" }
```
0xnm: so we expect configuration to be resolved during the configuration phase? well, if it works, then it is fine.

0xnm (dismissed):

tools/build-config/src/main/kotlin/com/datadog/gradle/plugin/transdeps/TransitiveDependenciesPlugin.kt
kikoveiga: Good point, switched it to `named(...).map` so the whole lookup remains lazy.

0xnm (approved):

tools/build-config/src/main/kotlin/com/datadog/gradle/plugin/transdeps/TransitiveDependenciesTask.kt
```diff
@@ -24,6 +24,9 @@ abstract class TransitiveDependenciesTask : DefaultTask() {
     @get:Input
     abstract val sortByName: Property<Boolean>
 
+    @get:Classpath
```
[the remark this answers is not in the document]
kikoveiga: This is a very edge case, no need to cover this.

